PR #56, opened 2020-03-10
.github: add Github actions to auto update Flatcar website docs

dongsupark:
Whenever a commit gets pushed to `github.com/flatcar-linux/docs`, the Github actions clones both `github.com/kinvolk/flatcar-website-docs` and `github.com/kinvolk/docs.flatcar-linux.org`, generate the docs, to finally push the final results to `github.com/kinvolk/docs.flatcar-linux.org`.


.github/workflows/flatcar-docs.yml
```diff
@@ -0,0 +1,19 @@
...
+    env:
+      GIT_SSH_COMMAND: "ssh -o StrictHostKeyChecking=no"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: auto sync flatcar-website-docs
+        env:
+          FLATCAR_DOCS_GITHUB_ACTIONS_KEY: ${{ secrets.FLATCAR_DOCS_GITHUB_ACTIONS_KEY }}
+          FLATCAR_WEBSITE_DOCS_GITHUB_ACTIONS_KEY: ${{ secrets.FLATCAR_WEBSITE_DOCS_GITHUB_ACTIONS_KEY }}
+        run: .github/workflows/sync-flatcar-docs.sh
```
alban: I was wondering whether to add the secret environment variables with a mask ([`::add-mask::{value}`](https://help.github.com/en/actions/reference/development-tools-for-github-actions#logging-commands)) for additional security (in case the script accidentally leaks it) but it seems GitHub already adds secrets automatically, so it's all good.

.github/workflows/sync-flatcar-docs.sh
```diff
@@ -0,0 +1,41 @@
...
+make copy-build
+
+popd || exit
+
+# Push the results into the remote git repo.
+pushd docs.flatcar-linux.org
+
+git commit -a -m "Update Flatcar docs ($(date +'%Y-%m-%d'))"
+git checkout -B site site
+git push origin site
```
pothos: I would rather like to have a review step here. I'm worried that the Python/Ruby library versions cause the formatting to break compared to the ones we use in the docker image.

dongsupark: Ok, let me see if it can be split out to another step for creating a pull request against `docs.flatcar-linux.org`.

dongsupark: Updated the PR.
So this PR does not run the syncing part any more. docs.flatcar-linux.org will run it instead.
This PR will simply send a repository dispatch event to docs.flatcar-linux.org, to trigger the actual build.

pothos: Thanks, that let's us do the regeneration manually if something looks off :)

PR #65, opened 2020-04-01
os: fix URLs to AWS cloudformation templates

dongsupark:
Now that AWS cloudformation has changed its URL scheme, we need to
adjust existing URLs to AWS cloudformation templates, like:

```
https://console.aws.amazon.com/cloudformation/home?region=eu-central-1#/stacks/new?stackName=flatcar-stable&templateURL=https://flatcar-prod-ami-import-eu-central-1.s3.amazonaws.com/dist/aws/flatcar-stable-hvm.template
```

### DIFF
--- a/os/booting-on-ec2.md
+++ b/os/booting-on-ec2.md
@@ -38,14 +38,14 @@ Flatcar Container Linux is designed to be updated automatically with different s
           <td rowspan="2">{{ region.name }}</td>
           <td class="dashed"><a href="http://aws.amazon.com/amazon-linux-ami/instance-type-matrix/">HVM (amd64)</a></td>
           <td class="dashed"><a href="https://console.{{ region_domain }}/ec2/home?region={{ region.name }}#launchAmi={{ region.hvm }}">{{ region.hvm }}</a></td>
-          <td class="dashed"><a href="https://console.{{ region_domain }}/cloudformation/home?region={{ region.name }}#cstack=sn%7EFlatcar-alpha%7Cturl%7Ehttps:%2F%2Fflatcar-prod-ami-import-eu-central-1.s3.amazonaws.com%2Fdist%2Faws%2Fflatcar-alpha-hvm.template" target="_blank"><img src="https://s3.amazonaws.com/cloudformation-examples/cloudformation-launch-stack.png" alt="Launch Stack"/></a></td>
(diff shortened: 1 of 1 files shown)


os/booting-on-ec2.md
```diff
@@ -38,14 +38,14 @@ Flatcar Container Linux is designed to be updated automatically with different s
           <td rowspan="2">{{ region.name }}</td>
           <td class="dashed"><a href="http://aws.amazon.com/amazon-linux-ami/instance-type-matrix/">HVM (amd64)</a></td>
           <td class="dashed"><a href="https://console.{{ region_domain }}/ec2/home?region={{ region.name }}#launchAmi={{ region.hvm }}">{{ region.hvm }}</a></td>
-          <td class="dashed"><a href="https://console.{{ region_domain }}/cloudformation/home?region={{ region.name }}#cstack=sn%7EFlatcar-alpha%7Cturl%7Ehttps:%2F%2Fflatcar-prod-ami-import-eu-central-1.s3.amazonaws.com%2Fdist%2Faws%2Fflatcar-alpha-hvm.template" target="_blank"><img src="https://s3.amazonaws.com/cloudformation-examples/cloudformation-launch-stack.png" alt="Launch Stack"/></a></td>
+          <td class="dashed"><a href="https://console.{{ region_domain }}/cloudformation/home?region={{ region.name }}#/stacks/new?stackName=Flatcar-alpha&templateURL=https://flatcar-prod-ami-import-eu-central-1.s3.amazonaws.com/dist/aws/flatcar-alpha-hvm.template" target="_blank"><img src="https://s3.amazonaws.com/cloudformation-examples/cloudformation-launch-stack.png" alt="Launch Stack"/></a></td>
```
alban, line 41: This domain name not seem to match the URL in the PR description. It should be either something like `eu-central-1.console.aws.amazon.com` (for that region) or just `console.aws.amazon.com` but I don't see how `console.{{ region_domain }}` works. Is the `console` prefix placed at the wrong part?

However, the previous `<td>` line has the same thing and the generated code does not seem to follow the template. How does it work?

pothos: `region_domain` is not about the AWS availability zone but about using `amazonaws.cn` for China.

pothos: The region is given via `?region=`  and AWS should redirect itself to `region.console`.

dongsupark: Right, `region_domain` is not the same as `region.name`. I had been puzzled by that. ;-)

PR #73, opened 2020-04-28
vmware: Update and expand the documentation

pothos:
- vmware: Update and expand the documentation    
    Important basic information was missing, like the usage of ovftool
    and how static IPs are configured.
    The relation of Ignition and coreos-cloudinit is explained because
    both use the same data source and Ignition has serveral limitations.
    Depending on the use case, some important behavior was documented
    such as which guestinfo variables result in what networkd setting
    when coreos-cloudinit is used, or how custom templating should be
    done with /run/metadata/coreos.
- os/provisioning: Mention the ct docker image

### DIFF
--- a/os/booting-on-vmware.md
+++ b/os/booting-on-vmware.md
@@ -13,6 +13,7 @@ Flatcar Container Linux is designed to be updated automatically with different s
     <li class="active"><a href="#stable" data-toggle="tab">Stable Channel</a></li>
     <li><a href="#beta" data-toggle="tab">Beta Channel</a></li>
     <li><a href="#alpha" data-toggle="tab">Alpha Channel</a></li>
+    <li><a href="#edge" data-toggle="tab">Edge Channel</a></li>
   </ul>
   <div class="tab-content coreos-docs-image-table">
     <div class="tab-pane active" id="stable">
@@ -33,25 +34,42 @@ Flatcar Container Linux is designed to be updated automatically with different s
       </div>
       <pre>curl -LO https://beta.release.flatcar-linux.net/amd64-usr/current/flatcar_production_vmware_ova.ova</pre>
     </div>
+    <div class="tab-pane" id="edge">
+      <div class="channel-info">
+        <p>The Edge channel includes bleeding-edge features with the newest versions of the Linux kernel, systemd and other core packages. Can be highly unstable. The current version is Flatcar Container Linux {{site.edge-channel}}.</p>
+      </div>
+      <pre>curl -LO https://edge.release.flatcar-linux.net/amd64-usr/current/flatcar_production_vmware_ova.ova</pre>
+    </div>
   </div>
 </div>
 
-### Booting with VMware ESXi
+### Booting with VMware vSphere/ESXi from the web interface
 
-Use the vSphere Client to deploy the VM as follows:
+Use the vSphere Client/ESXi web interface to deploy the VM as follows:
 
 1. In the menu, click `File` > `Deploy OVF Template...`
 2. In the wizard, specify the location of the OVA file downloaded earlier
 3. Name your VM
 4. Choose "thin provision" for the disk format
-5. Choose your network settings
+5. Choose your network settings and [specify provisioning userdata][guestinfo]
 6. Confirm the settings, then click "Finish"
 
 Uncheck `Power on after deployment` in order to edit the VM before booting it the first time.
 
 The last step uploads the files to the ESXi datastore and registers the new VM. You can now tweak VM settings, then power it on.
 
+### Booting with VMware vSphere/ESXi from the command line with ovftool
+
+Use the [`ovftool`][ovftool] to deploy from the command line as follows:
+
+```sh
+$ ovftool --name=testvm --skipManifestCheck --noSSLVerify --datastore=datastore1 --powerOn=True --net:"VM Network=VM Network" --X:waitForIp --overwrite --powerOffTarget --X:guest:coreos.config.data=$(cat ignition_config.json | base64 --wrap=0) --X:guest:coreos.config.data.encoding=base64 ./flatcar_production_vmware_ova.ova 'vi:///<YOUR_USER>:<ESXI_PASSWORD>@<ESXI_HOST_IP>'
+```
+
+This assumes that you downloaded `flatcar_production_vmware_ova.ova` to your current folder, and that you want to specify an Ignition config as userdata from `ignition_config.json`.
+
 *NB: These instructions were tested with an ESXi v5.5 host.*
+
 
 ### Booting with VMware Workstation 12 or VMware Fusion
 
@@ -61,9 +79,8 @@ Run VMware Workstation GUI:
 2. In the wizard, specify the location of the OVA template downloaded earlier
 3. Name your VM, then click `Import`
 4. (Press `Retry` *if* VMware Workstation raises an "OVF specification" warning)
-5. Edit VM settings if necessary
-6. [Modify the `.vmx` file][guestinfo] to pass an Ignition config containing at least one valid SSH key
-7. Start your Flatcar Container Linux VM
+5. Edit VM settings if necessary and [specify provisioning userdata][guestinfo]
+6. Start your Flatcar Container Linux VM
 
 *NB: These instructions were tested with a Fusion 8.1 host.*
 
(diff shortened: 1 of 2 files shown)


t-lo:
Would it make sense to also update the names from `guestinfo.coreos.*` to `guestinfo.ignition.*` in the documentation or is further work (e.g. in cloud-init) required before we can do that?

os/booting-on-vmware.md
```diff
@@ -140,30 +246,29 @@ guestinfo.coreos.config.data = "ewogICJpZ25pdGlvbiI6IHsgInZlcnNpb24iOiAiMi4wLjAi
...
 
-## Logging in
+Instead of providing the userdata inline, you can also specify a remote HTTP location in `guestinfo.coreos.config.url`.
+Both Ignition and coreos-cloudinit support it but Ignition relies on DHCP in the initramfs which means that it can't fetch remote resources if you have to use static IPs.
 
-Networking can take some time to start under VMware. Once it does, press enter a few times at the login prompt and you should see an IP address printed on the console:
+## Logging in
 
-![VMware IP Address](img/vmware-ip.png)
+The VGA console will have autologin enabled for releases after 2020/04/28.
```
blixtra, line 262: Why the date here? It's April 28th now so why not just state it?

os/booting-on-vmware.md
```diff
@@ -140,30 +246,29 @@ guestinfo.coreos.config.data = "ewogICJpZ25pdGlvbiI6IHsgInZlcnNpb24iOiAiMi4wLjAi
...
 
-In this case the IP is `10.0.1.81`.
+Networking can take some time to start under VMware. Once it does, you will see the IP when typing `ip a` or in the VM info that VMware displays.
 
-Now you can login to the host at that IP using your SSH key, or the password set in your cloud-config:
+You can login to the host at that IP using your SSH key, or the password set in your cloud-config:
 
 ```sh
-ssh core@10.0.1.81
+ssh core@YOURIP
```
blixtra, line 269: We should probably converge on a convention for these substitutions. But fine for now.

os/booting-on-vmware.md
```diff
@@ -73,31 +88,117 @@ Flatcar Container Linux can also be installed by booting the virtual machine via
 
 ## Container Linux Configs
 
-Flatcar Container Linux allows you to configure machine parameters, configure networking, launch systemd units on startup, and more via Container Linux Configs. These configs are then transpiled into Ignition configs and given to booting machines. Head over to the [docs to learn about the supported features][cl-configs].
+Flatcar Container Linux allows you to configure machine parameters, configure networking, launch systemd units on startup, and more via Container Linux Configs. These configs are then [transpiled][transpiler] into Ignition configs and given to booting machines. Head over to the [docs to learn about the supported features][cl-configs].
```
blixtra, line 91: Should this not be `[transpiler][transpiler]`? I don't see `transpiled` below.

os/booting-on-vmware.md
pothos: It's a way of publishing a statement without synchronizing it with the release date. We haven't done a release yet and if we do, people could check if the channel they run is on a version released after this date or not. I would remove this when Stable is released but afterwards it wouldn't hurt either. Something like _future/next release_ is unclear as long as I don't know when the doc change was done.

os/booting-on-vmware.md
pothos: `[transpiled]` is the word in the text, linking to `[transpiler]` defined at the end of the document.

os/booting-on-vmware.md
```diff
@@ -33,26 +34,41 @@ Flatcar Container Linux is designed to be updated automatically with different s
...
 4. Choose "thin provision" for the disk format
-5. Choose your network settings
+5. Choose your network settings and [specify provisioning userdata][guestinfo]
 6. Confirm the settings, then click "Finish"
 
 Uncheck `Power on after deployment` in order to edit the VM before booting it the first time.
 
 The last step uploads the files to the ESXi datastore and registers the new VM. You can now tweak VM settings, then power it on.
 
+Or use the [`ovftool`][ovftool] to deploy from the command line as follows:
```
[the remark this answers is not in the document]
pothos: Sure, that's better. I changed it.

pothos:
`guestinfo.ignition.*` support is not released yet on any channel, so I kept the old names.

os/booting-on-vmware.md
```diff
@@ -106,6 +209,10 @@ The VMware guestinfo interface is a mechanism for VM configuration. Guestinfo pr
 
 * Configure guestinfo in the OVF for deployment. Software like [vcloud director][vcloud director] manipulates OVF descriptors for guest configuration. For details, check out this VMware blog post about [Self-Configuration and the OVF Environment][ovf-selfconfig].
 
+* The ESXi web UI and VMware Workstation Player display the OVF guestinfo variables or allow to add them as parameters in the VM settings before deployment. The can also be changed and added later (but for Ignition configs that requires `touch /boot/flatcar/first_boot` so that Ignition runs again on the next boot).
```
marga-kinvolk, line 212: should this be "and allow to add them"?
Also the next sentence has a typo: s/The can/They can/

os/booting-on-vmware.md
```diff
@@ -122,16 +229,17 @@ vmware-cmd /vmfs/volumes/[...]/<VMNAME>/<VMNAME>.vmx setguestinfo guestinfo.<pro
 
 Guestinfo configuration set via the VMware API or with `vmtoolsd` from within the Flatcar Container Linux guest itself are stored in VM process memory and are lost on VM shutdown or reboot.
 
-### Defining the Ignition config in Guestinfo
+### Defining the Ignition config or coreos-cloudinit Cloud-Config in Guestinfo
 
-If the `guestinfo.coreos.config.data` property is set, Ignition will apply the referenced config on first boot.
+If the `guestinfo.coreos.config.data` or `guestinfo.coreos.config.url` userdata property contains an Ignition config, Ignition will apply the referenced config on first boot during the initramfs phase. If it contains a Cloud-Config or script, Ignition will enable a service for coreos-cloudinit that will run on every boot and apply the config.
```
marga-kinvolk, line 234: What happens if both variables are set?
In any case, I think this should say something like "If either the ... or ... userdata properties contain ..."

os/booting-on-vmware.md
```diff
@@ -140,30 +248,29 @@ guestinfo.coreos.config.data = "ewogICJpZ25pdGlvbiI6IHsgInZlcnNpb24iOiAiMi4wLjAi
 guestinfo.coreos.config.data.encoding = "base64"
 ```
 
-This example will be decoded into:
+This example will be decoded into the follwoing Ignition config, but a Cloud-Config can be specified the same way in the variable:
```
marga-kinvolk, line 251: s/follwoing/following/

os/booting-on-vmware.md
marga-kinvolk: Maybe include the release number instead of the date?

os/booting-on-vmware.md
pothos: Thanks

os/booting-on-vmware.md
pothos: `.data` takes preference but I now used _either or_.

os/booting-on-vmware.md
pothos: Thanks.
Added _either or_ because when they are not displayed automatically then they can be added as parameters in the settiings.

os/booting-on-vmware.md
```diff
@@ -73,31 +90,117 @@ Flatcar Container Linux can also be installed by booting the virtual machine via
...
+  listen_peer_urls:            "http://10.0.0.2:2380"
   # generate a new token for each unique cluster from https://discovery.etcd.io/new?size=3
   # specify the initial size of your cluster with ?size=X
   discovery:                   "https://discovery.etcd.io/<token>"
 ```
 
+For DHCP you don't need to specify any networkd units.
+
+After transpilation, the resulting JSON content can be used in `guestinfo.coreos.config.data` after encoding it to base64 and setting `guestinfo.coreos.config.data.encoding` to `base64`.
+If DHCP is used the JSON file can also be uploaded to a web server and fetched by Ignition if the HTTP(s) URL is given in `guestinfo.coreos.config.url`.
```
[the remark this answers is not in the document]
pothos: Done

os/booting-on-vmware.md
pothos: Hm, the last release number per channel where it is not included? That could work.

PR #76, opened 2020-04-29
Ignition: Document how to enable systemd services when reprovisioning

pothos:
This question came up on IRC and since systemd presets are not mentioned
and resetting the presets is also not easy, just document how a symlink
can be used to ensure that a service is enabled.


pothos:
I think it's ok to use both for short examples. In general, we need something like this https://coreos.com/os/docs/latest/adding-users.html where both are displayed besides each other (tracked in https://github.com/flatcar-linux/Flatcar/issues/72)

PR #82, opened 2020-05-07
installing to disk: this data is yaml, not json

vbatts:

# It's YAML, not JSON

and while the installer does not pay attention to the file extension, the example should be consistent.

# How to use

validate that putting the cluster config YAML into a filename ending in .yaml (rather than .json) still works

# Testing done

I did an install using a `*.yaml` file.


pothos:
It's ok to close, or?

PR #86, opened 2020-05-15
update_engine_client requires sudo access

chasebolt:
after running the update from coreos to flatcar, it failed as `update_engine_client` was requiring root access


pothos:
Normally that is not the case. Maybe your user wasn't `core` or something else in your setup?

Thanks for the PR. We can add it since it doesn't hurt. Will have a look on Monday.

chasebolt:
Correct, I have my own user defined and do not use `core`. I assumed that was the reason as well.

PR #87, opened 2020-05-18
Note that update-engine needs to run after the reboot

pothos:
The documentation explained how to disable update-engine permanently
but the fact that updates don't work if update-engine is running only
shortly to update was not called out.
Explain that update-engine needs to run after the reboot, too, and show
how to check that the service can be disabled again safely.

### DIFF
--- a/os/update-strategies.md
+++ b/os/update-strategies.md
@@ -81,6 +81,22 @@ systemd:
       mask: true
 ```
 
+**Note:** If you want to manually trigger an update after having masked `update-engine`,
+you'll need to unmask the service, start `update-engine` to trigger an update, and
+**keep the service unmasked** until the next reboot is completed and `update-engine` started
+and marked the updated partition as successful.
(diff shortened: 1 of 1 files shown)


os/update-strategies.md
```diff
@@ -81,6 +81,20 @@ systemd:
       mask: true
 ```
 
+**Note:** If you temporarily unmask and start `update-engine` to trigger an update,
```
marga-kinvolk, line 84: I think I would phrase this a little differently:

**Note**: If you want to manually trigger an update after having masked `update-engine`, you'll need to unmask the service and **keep it unmasked** until the next reboot is completed and `update-engine` has marked the updated partition as successful. Otherwise, the update will be considered unsuccessful and all following reboots will use the old partition again.

pothos: Thanks, have taken it but with slightly more detail.

PR #89, opened 2020-05-18
os: flatcar name and shell cleanup

vbatts:

# flatcar name and shell cleanup

overly pedantic shell cleanup and new path

# How to use

a flatcar instance

# Testing done

walked through the examples.


vbatts:
edited because the `flatcar-sources` also has the 404, but no overlays

vbatts:
it appears since there are coreos.com URLs posted in the `make.conf` for `PORTAGE_BINHOST=`, this command is attempting to get that binary build.

PR #91, opened 2020-06-05
toolbox: Document how to run tmux in the background and attach to it

pothos:
The toolbox can only run once and it's not easy to run a background
process or share a session.
Document how to do that with some tricks.

# How to use

See commands.

# Testing done

Tested in a VM with multiple SSH connections that attach to the tmux session.

### DIFF
--- a/os/install-debugging-tools.md
+++ b/os/install-debugging-tools.md
@@ -62,6 +62,39 @@ This means two important things:
 * The container filesystem will take up space on disk (a few hundred MiB
 for the default `fedora` container)
 
+## Spawn a toolbox with tmux in the background
+
+Since `toolbox` can only be started once it is not straightforward to use `tmux`
+for long-running jobs or sharing a debugging session with someone else.
+
+To keep user processes running in the background after logging out with SSH,
+you need to start them via `systemd-run` because _process lingering_ is disabled
+by default in logind and all non-service user processes are killed on logout.
+Spawn a user service to persist the toolbox container with the `tmux` process
+even when you log out with SSH.
+The following command line will ensure `tmux`, `strace` and `pidof` are installed
+in the container, then create a new `tmux` session to which you can later attach,
+and keep the service active by waiting with `strace` until the `tmux` process exits.
+
+```sh
+systemd-run --user toolbox sh -c 'dnf install -y tmux strace procps-ng; tmux new-session -d -s sharedsession; strace -p "$(pidof tmux)"'
+```
+
+With `-d` we tell `tmux` to not allocate a TTY now (needed for `systemd-run`) but run a
+new session in the background.
+Because `tmux` forks away, we cannot use `wait` in the shell to wait for children but need
+to use `strace` to have a foreground process running that prevents `toolbox` from quitting.
+
+Once this is running you can can attach to the `tmux` session as often as you want from any SSH connection.
+
+```sh
+sudo nsenter -t "$(pidof tmux | cut -d ' ' -f 1)" -a tmux a
+```
+
+As usual with `tmux` you can attach and detach to the session as many times as you want because detaching
+still keeps `tmux` running in the background. But keep in mind that if you exit the session, the process
+started with `systemd-run` will terminate and you'll have to start the service again with `systemd-run`.
+
 ## SSH directly into a toolbox
 
 Advanced users can SSH directly into a toolbox by setting up an `/etc/passwd` entry:


os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
 * The container filesystem will take up space on disk (a few hundred MiB
 for the default `fedora` container)
 
+## Spawn a toolbox with tmux in the background
+
+Since `toolbox` can only be started once it is not straightforward to use `tmux`
+for long-running jobs or sharing a debug session with someone else.
```
marga-kinvolk, line 68: s/debug/debugging/

os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
...
+
+Lingering processes are disabled by default which means that you have to start the
```
marga-kinvolk, line 70: This sentence is confusing. What does it mean that lingering processes are disabled?

Also, I think it would be clearer to start with the goal rather than with the blocker:

"To keep background processes running inside Toolbox, you need to start them via `systemd-run`. That way, they will persist even if you log out".

I'm not sure what you mean by logout with SSH, I thought this was about toolbox, not ssh?

os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
...
+background process via `systemd-run` to have `tmux` persist when you log out with SSH.
+
+```sh
+systemd-run --user toolbox sh -c 'dnf install -y tmux strace procps-ng; tmux new-session -d -s sharedsession; strace -p "$(pidof tmux)"'
```
marga-kinvolk, line 74: This command line is lacking some intro about what you're about to paste. Something like:

"The following command line will ensure tmux, strace and pidof are installed in the container and then create a new tmux session to which you can later attach."

I know you explain the parameters later, but giving the line a bit of an intro makes it easier to understand what's going on.

marga-kinvolk, line 74: The command line is very long, which means that when rendered one needs to use the scrollbar to see it fully.  Can you break it into 2 or more lines so that it fits?

os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
...
+```
+
+With `-d` you tell `tmux` to not allocate a TTY now (needed for `systemd-run`) but run a
```
marga-kinvolk, line 77: I'd switch all this paragraph to "we" instead of you. After all, you're the one that came up with this command line, it's kinda weird that you're explaining it to me in second person.

os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
...
+Because `tmux` forks away, you cannot use `wait` in the shell to wait for children but need
+to use `strace` to have a foreground process running that prevents `toolbox` from quitting.
+
+Now you can can attach to the `tmux` session as often as you want from any SSH connection.
+
+```sh
+sudo nsenter -t "$(pidof tmux | cut -d ' ' -f 1)" -a tmux a
+```
+
+Detaching leaves it run in the background but if you exit the session, the systemd
```
marga-kinvolk, line 88: s/it run/it running/

os/install-debugging-tools.md
```diff
@@ -62,6 +62,32 @@ This means two important things:
...
+```sh
+systemd-run --user toolbox sh -c 'dnf install -y tmux strace procps-ng; tmux new-session -d -s sharedsession; strace -p "$(pidof tmux)"'
+```
+
+With `-d` you tell `tmux` to not allocate a TTY now (needed for `systemd-run`) but run a
+new session in the background.
+Because `tmux` forks away, you cannot use `wait` in the shell to wait for children but need
+to use `strace` to have a foreground process running that prevents `toolbox` from quitting.
+
+Now you can can attach to the `tmux` session as often as you want from any SSH connection.
```
marga-kinvolk, line 82: s/Now/Once this is running,/

os/install-debugging-tools.md
marga-kinvolk: Also, I'd be a bit more verbose: "You can attach and detach to the session as many times as you want because detaching still keeps the process running in the background. But keep in mind that if you exit the session, the process started with `systemd-run` will terminate and you'll have to run it again.

os/install-debugging-tools.md
pothos: Thanks for the comments, will try to add the missing information. Any user process, including the whole toolbox container, is normally killed when the SSH session ends.

os/install-debugging-tools.md
pothos: Done

os/install-debugging-tools.md
pothos: Thanks

os/install-debugging-tools.md
pothos: Done

os/install-debugging-tools.md
pothos: Inserting explicit line breaks makes it difficult to copy the command. As far as I know the rendering of the doc websites will break the lines without explicit line breaks.

pothos: Thanks, included it.

os/install-debugging-tools.md
pothos: Done

os/install-debugging-tools.md
pothos: Done

os/install-debugging-tools.md
```diff
@@ -62,6 +62,39 @@ This means two important things:
...
+## Spawn a toolbox with tmux in the background
+
+Since `toolbox` can only be started once it is not straightforward to use `tmux`
+for long-running jobs or sharing a debugging session with someone else.
+
+To keep user processes running in the background after logging out with SSH,
+you need to start them via `systemd-run` because _process lingering_ is disabled
+by default in logind and all non-service user processes are killed on logout.
+Spawn a user service to persist the toolbox container with the `tmux` process
+even when you log out with SSH. The following command line will ensure `tmux`,
```
marga-kinvolk, line 74: NIt pick: I'd add a line break after the period

os/install-debugging-tools.md
marga-kinvolk: Well, I meant with \ in between to allow the copy & paste to work.

os/install-debugging-tools.md
pothos: Done

os/install-debugging-tools.md
pothos: Pasting works but only the last line can be edited and I find it very confusing how readline handles this.

PR #100, opened 2020-07-16
Document caveat with SDK package URLs

pothos:
See https://github.com/flatcar-linux/scripts/pull/80

### DIFF
--- a/os/sdk-modifying-flatcar.md
+++ b/os/sdk-modifying-flatcar.md
@@ -114,19 +114,22 @@ to switch to a recent nightly version because it only dereferences the nightly o
 build environment reproducible.
 
 ```
-./set_version --dev-board --board-version amd64-usr/CHANNEL-nightly
+./set_version --dev-board --board-version amd64-usr/CHANNEL-nightly --no-dev-sdk
+# --no-dev-sdk needed because --dev-board implies --dev-sdk
 ```
 
-To switch to a Flatcar release use the following:
+To switch to a Flatcar release when you checked out release branches use the following:
 
 ```
-./set_version --no-dev-board --board-version x.y.z  # --no-dev-board disables any previous --dev-board setting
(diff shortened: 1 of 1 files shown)


os/sdk-modifying-flatcar.md
```diff
@@ -114,19 +114,21 @@ to switch to a recent nightly version because it only dereferences the nightly o
 build environment reproducible.
 
 ```
-./set_version --dev-board --board-version amd64-usr/CHANNEL-nightly
+./set_version --dev-board --board-version amd64-usr/CHANNEL-nightly --no-dev-sdk
+# --no-dev-sdk needed because --dev-board implies --dev-sdk
 ```
 
 To switch to a Flatcar release use the following:
```
marga-kinvolk, line 121: Can you add a bit more text here explaining why someone would want to switch?

os/sdk-modifying-flatcar.md
```diff
@@ -114,19 +114,21 @@ to switch to a recent nightly version because it only dereferences the nightly o
...
 
 ```
-./set_version --no-dev-board --board-version x.y.z  # --no-dev-board disables any previous --dev-board setting
+./set_version --no-dev-board --board-version x.y.z
+# --no-dev-board disables any previous --dev-board setting
 ```
 
 You should also use this command to switch to a newer SDK release if available.
```
marga-kinvolk, line 128: I find the "should" here confusing... Shouldn't this be "can"? As in: "You can also use this command"... ?

pothos: Thanks, changed and added _so that you don't have to rebuild SDK packages when you checked out more recent branches._

os/sdk-modifying-flatcar.md
pothos: Thanks, added _when you checked out release branches_ – does this help?

PR #102, opened 2020-07-20
os/sdk-modifying-flatcar.md: Document cork create usage

pothos:
The default branches to be checked out will be the Alpha development
branches with the latest Alpha SDK.
See https://github.com/flatcar-linux/manifest/pull/14
The release branch of the manifest has changed from flatcar-build-x.y.z
to flatcar-CHANNEL-x.y.z which is updated here as well.
As replacement mechanism for the current CHANNEL-master.xml files
a new maintenance.xml is proposed that is part of the release tag/
branch of the manifest repository (not yet released).

### DIFF
--- a/os/sdk-modifying-flatcar.md
+++ b/os/sdk-modifying-flatcar.md
@@ -73,12 +73,15 @@ cork create # This will request root permisions via sudo
 cork enter # This will request root permisions via sudo
 ```
 
-By default `cork create` will download the latest SDK used for the Stable release which can be changed with `--sdk-version x.y.z`.
+By default `cork create` will download the latest SDK used for the Alpha release which can be changed with `--sdk-version x.y.z`.
 You can also change the SDK release later.
 
-A manifest defines which source code branches are checked out. The default manifest uses the development branches for the Stable channel.
-If you want to patch a specific release, you can pass `--manifest-branch flatcar-build-x.y.z` and if you want to work on a channel other
-than Stable you can pass `--manifest-name CHANNEL-master.xml`. You can also change the branches later with git.
+A manifest defines which source code branches are checked out. The default manifest uses the development branches for the next Alpha major version.
+If you want to check out a specific release, you can pass `--manifest-branch flatcar-CHANNEL-x.y.z`.
+If you want to add changes to a major version that is propagated through the channels and gets maintenance releases increasing the minor version,
+you need to use the maintenance branch `flatcar-MAJORVERSION` (the major version is the _x_ in _x.y.z_) which you can check out by passing
+`--manifest-branch flatcar-CHANNEL-x.y.z --manifest-name maintenance.xml` for any release of that major version family (not for releases <= 2513.2.0).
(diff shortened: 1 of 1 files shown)


pothos:
The `maintenance` manifest is there to work on a particular maintenance branch of a release. When preparing a PR to fix something for a release, one needs to check out the maintenance branch. Using the release manifest would end up checking out the exact release tags, which does not show the current development state of that maintenance branch and also doesn't tell where to file the PRs.

marga-kinvolk:
Ok, I think I understand this now, but it's just that "maintenance" conveys the exact opposite message for me (something is in maintenance mode when it's no longer in development mode). Would you be open to giving it another name? "development" would work for me.

pothos:
It's not really where development takes place but rather maintenance, that's why I chose the name but a better name is welcome. I think `development` fits well as a name for the Alpha development manifest.

pothos:
This here mentions maintenance branches https://git-scm.com/docs/gitworkflows
We could call them release branches but to me this conflicts with the release manifest which pins specific commits.

marga-kinvolk:
I think what makes it different in our case is how releases graduate from one channel to the other.  So, moving a release from 1234.1.0 to 1234.2.0 is not really "maintenance" but rather a channel promotion... Anyway, I've been trying to come up with something better and none of the options I can think are really good (ongoing, major-head, branch-head, release-next).  Can you at least explain in the doc a bit more what you mean with "maintenance" in this case?

os/sdk-modifying-flatcar.md
```diff
@@ -73,12 +73,13 @@ cork create # This will request root permisions via sudo
...
 
-By default `cork create` will download the latest SDK used for the Stable release which can be changed with `--sdk-version x.y.z`.
+By default `cork create` will download the latest SDK used for the Alpha release which can be changed with `--sdk-version x.y.z`.
 You can also change the SDK release later.
 
-A manifest defines which source code branches are checked out. The default manifest uses the development branches for the Stable channel.
-If you want to patch a specific release, you can pass `--manifest-branch flatcar-build-x.y.z` and if you want to work on a channel other
-than Stable you can pass `--manifest-name CHANNEL-master.xml`. You can also change the branches later with git.
+A manifest defines which source code branches are checked out. The default manifest uses the development branches for the next Alpha major version.
+If you want to patch a specific release, you can pass `--manifest-branch flatcar-CHANNEL-x.y.z` and if you want to work on the maintenance branch
```
marga-kinvolk, line 80: What does "patch" mean in this context? Maybe "check out" would be more accurate?

Maybe the second part could be: "and if you want to add changes to a major version, you need to use the `maintenance.xml` file for a release in that family" ?

pothos: Yes, patch was wrong because it fits only to the second part of adding changes.
I took your suggestion with a few more words as separate sentence. Since the user never interacts with the actual file, I focused on showing which command to use.

pothos:
All good :)

PR #114, opened 2020-09-04
Add AppArmor step and backing_fmt parameter

gomex:
# Add AppArmor step and backing_fmt parameter

AppArmor step is necessary to run this doc on ubuntu
Backing_fmt parameter  is necessary

### DIFF
--- a/os/booting-with-libvirt.md
+++ b/os/booting-with-libvirt.md
@@ -108,7 +108,7 @@ restorecon -R "/var/lib/libvirt/flatcar-linux/flatcar-linux1"
 
 If the host uses AppArmor, allow `qemu` to access the config files:
 
-```sh
+```shell
 echo "  # For ignition files" >> /etc/apparmor.d/abstractions/libvirt-qemu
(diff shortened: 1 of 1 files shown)


os/booting-with-libvirt.md
```diff
@@ -106,6 +106,13 @@ semanage fcontext -a -t virt_content_t "/var/lib/libvirt/flatcar-linux/flatcar-l
 restorecon -R "/var/lib/libvirt/flatcar-linux/flatcar-linux1"
 ```
 
+If the host uses AppArmor, allow the qemu access to the config:
```
marga-kinvolk, line 109:
```suggestion
If the host uses AppArmor, allow `qemu` to access the config files:
```

os/booting-with-libvirt.md
```diff
@@ -106,6 +106,13 @@ semanage fcontext -a -t virt_content_t "/var/lib/libvirt/flatcar-linux/flatcar-l
...
+
+```sh
```
vbatts, line 111:
```suggestion
```shell
```

PR #119, opened 2020-10-30
os/*.md: consistency and markdown lint

vbatts:
Across the docs, there were inconsistent use of code block languages,
duplicate heading (for anchoring), and markdown cleanup.

We still have a fair amount of inline HTML. I changed some of the "more
information" buttons, into a simple list. I think this is more portable.


dongsupark (approved):
Thanks!

PR #124, opened 2020-11-13
New docs

joaquimrocha:
# New structure and format for docs

This is the structure that can be consumed by the new docs portal.


docs/bare-metal/booting-with-ipxe.md
```diff
@@ -73,7 +76,7 @@ initrd ${base-url}/flatcar_production_pxe_image.cpio.gz
 boot</pre>
     </div>
     <div class="tab-pane active" id="stable-create">
-      <p>The Stable channel should be used by production clusters. Versions of Flatcar Container Linux are battle-tested within the Beta and Alpha channels before being promoted. The current version is Flatcar Container Linux {{site.stable-channel}}.</p>
+      <p>The Stable channel should be used by production clusters. Versions of Flatcar Container Linux are battle-tested within the Beta and Alpha channels before being promoted. The current version is Flatcar Container Linux {{< param stable_channel >}}.</p>
```
vbatts, line 79: i'm not sure where this param comes from, but 
Screenshot from 2020-11-13 11-54-18
it seems to be out of date. Not a blocker, but something to follow-up on

docs/_index.md
```diff
@@ -0,0 +1,165 @@
+---
+content_type: flatcar
+title: Flatcar Container Linux
+main_menu: true
+weight: 40
+cascade:
+  alpha_channel: 2605.0.0
+  stable_channel: 2512.3.0
```
vbatts, line 8: oh, maybe those params are inherited from here?

docs/_index.md
```diff
@@ -0,0 +1,165 @@
+---
```
vbatts, line 1: So, until we finish the redirect of docs.flatcar-linux.org, then we'll have two index files to keep in lock step?

vbatts (approved):
OHMAN
a couple of things to follow-up on, but this is huge. Thanks for the effort to get these docs unified.

vbatts:
this will shortly get a new build PR for the netlify deployment

PR #128, opened 2020-11-23
.github: Remove old GitHub Action

pothos:


marga-kinvolk (approved):
LGTM.

But I'm guessing this one should also wait until the redirect is in place?

PR #129, opened 2020-11-23
Remove old index.md

pothos:
The new index is under docs/_index.md.


marga-kinvolk (commented):
I don't know how the new docs work, but in the other PR there was a comment from Vincent saying that both indexes needed to exist until the redirection was in place, and the redirection doesn't seem to be in place?

pothos:
Ah, so this is also for github? Then I will update this PR to include a missing link.

pothos:
Hm, but where should these links point to? They look broken.

PR #131, opened 2020-11-27
Add an example to set the hostname from metadata

dr4Ke:
# Add an example to set the hostname from metadata

It was not obvious to me how I could set the hostname of an openstack instance. The coreos-metadata service doesn't seem to do it by itself, as I would have expected.

So I searched how to do it with ignition and it took me too much time. I thought that a mention in the documentation could help.

# How to use

N/A

# Testing done

This configuration was tested with success on an openstack cloud.

### DIFF
--- a/docs/reference/migrating-to-clcs/_index.md
+++ b/docs/reference/migrating-to-clcs/_index.md
@@ -251,7 +251,7 @@ storage:
         inline: coreos1
 ```
 
-If your cloud provider uses a meta-data service, you can get the hostname from it. For example with openstack:
+On a cloud having an `http` metadata service, the hostname can be retrieved automatically. For example with openstack:
 
 ```yaml
 storage:
--- a/docs/reference/migrating-to-clcs/_index.md
+++ b/docs/reference/migrating-to-clcs/_index.md
@@ -251,6 +251,17 @@ storage:
         inline: coreos1
 ```
 
+If your cloud provider uses a meta-data service, you can get the hostname from it. For example with openstack:
+
+```yaml
+storage:
+  files:
+    - path:       "/etc/hostname"
+      contents:
+        remote:
+          url: http://169.254.169.254/latest/meta-data/hostname
+```
+
 ### users
 
 The `users` section in a cloud-config can be used to add users and specify many properties about them, from groups the user should be in to what the user's shell should be.


docs/reference/migrating-to-clcs/_index.md
```diff
@@ -251,6 +251,17 @@ storage:
         inline: coreos1
 ```
 
+On a cloud having an `http` metadata service, the hostname can be retrieved automatically. For example with openstack:
```
marga-kinvolk, line 254:
```suggestion
The hostname can also be retrieved from an `http` metadata service, if available. For example, on OpenStack, which includes a metadata service available on 169.254.169.254, you could use:
```

docs/reference/migrating-to-clcs/_index.md
```diff
@@ -251,6 +251,17 @@ storage:
...
+
+```yaml
+storage:
+  files:
+    - path:       "/etc/hostname"
```
marga-kinvolk, line 259:
```suggestion
    - path: "/etc/hostname"
```

marga-kinvolk (commented):
Thanks! I've added a suggestion to reword the intro a bit, to generalize the issue and make the magic IP clear for people not familiar with openstack (so that it's easier to translate to other clouds).

Also, the path variable doesn't need all those spaces if the `filesystem` field is not there. Although, maybe it should? I'm not sure when it should be there or not...

pothos (commented):
On some platforms this is done by afterburn (coreos-metadata) in the initramfs. We can also try to do this on every platform on the first boot before Ignition runs.

docs/reference/migrating-to-clcs/_index.md
invidian: Also quotes could be dropped OR be consistent with `url:`.

invidian: And `mode` must be specified, otherwise strict parsing will fail.

pothos:
Here is what needs to be done to make it work out of the box instead of adding docs:

Similar to https://github.com/kinvolk/bootengine/blob/flatcar-master/dracut/30ignition/flatcar-digitalocean-network.service a service for ec2 should be added to bootengine, but without the `--network-units=/run/systemd/network/` parameter.
And then in this ec2 condition here https://github.com/kinvolk/bootengine/blob/flatcar-master/dracut/30ignition/ignition-generator#L51 one would add `add_requires SERVICENAME`.

pothos:
Created a PR here: https://github.com/kinvolk/bootengine/pull/21

pothos:
Setting the hostname is now merged and will be part of the next major version release in the Alpha channel.
Thanks again for the docs PR!

PR #135, opened 2020-12-07
sdk-disk-partitions.md: Correct OEM partition mount description

pothos:
The OEM partition is not mounted read-only but can be written to.


marga-kinvolk (approved):
LGTM

PR #137, opened 2020-12-09
Add docs for Flatcar Pro on Azure

pothos:
This documents the Flatcar Pro marketplace images and how to query them
    from the CLI. Currently the latest release does not support the GPU
    drivers and an old version has to be used, this is documented as well.


sayanchowdhury (approved):

t-lo (changes requested):
I'd rather not publish the Pro Beta part (but instead release a new Azure image that has updates disabled).
Main reason here is that we involuntarily publish a way for "free" offer users to get Pro updates.

pothos:
Updated

PR #138, opened 2020-12-10
Using a custom Docker version

pothos:
- Add new docs on deploying a custom Docker version through Ignition
    Even though Torcx exists for some time already, it's not really
    accessible and has several flaws. Therefore, users tend to prefer to
    provisioning custom binaries through Ignition but there are no docs
    yet about this.
    Document how to replicate the Docker setup Flatcar has by writing new
    files to /etc and /opt/bin as a starting point for users to do their
    customizations, like just downloading containerd or additional tools.
- customizing-docker: Show how to use a custom containerd configuration
- docs/container-runtimes/customizing-docker.md: Capital D for Docker

# How to use/testing done

Save the config file and convert it: `cat config.yaml | container-linux-config-transpiler/bin/ct > config.json`
Start a VM: `./flatcar_production_qemu.sh -i config.json`
Log in: `ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 2222 core@127.0.0.1`, then run the `export` command stated there and check that Docker works.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -80,6 +80,7 @@ Docker |
 -------------- |
 [Getting started with Docker][docker] |
 [Customizing Docker][customizing-docker] |
+[Use a custom Docker or containerd version][use-a-custom-docker-or-containerd-version] |
 [Switching from Docker to containerd for Kubernetes][containerd-for-kubernetes] |
 
 ### Reference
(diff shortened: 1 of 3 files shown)


t-lo (approved):
If there's some overlap with https://github.com/kinvolk/flatcar-docs/blob/main/docs/container-runtimes/switching-from-docker-to-containerd-for-kubernetes.md then we might want to reference the kubernetes containerd guide, otherwise LGTM, thank you!

pothos:
Yes, added a link to it from the _Customizing Docker_ page.

PR #141, opened 2020-12-22
Rearrange sections and provide aliases

marga-kinvolk:
# Rearrange sections and provide aliases

The changes included rearrange the left bar navigation, both for the main section and for the cloud-providers section.

Additionally, it includes automatically generated aliases from the old paths.  Files that were renamed or that were not in the os section won't work, but this will cover most of them.

# How to use / Testing done

I've generated the website myself and verified that the order was correct and that accessing a page like http://localhost:1313/docs/flatcar-container-linux/mypreviewversion/os/booting-on-azure.md redirected to http://localhost:1313/docs/flatcar-container-linux/mypreviewversion/cloud-providers/booting-on-azure/

I intend to also send a PR for the website to re-add the redirection from the old address.


marga-kinvolk:
I've done a few more checks regarding files that were renamed or had different paths than the obvious, and I added any necessary redirections.

I think this is now covering all pages. I also fixed the ignition files that didn't need an alias because they were already in a separate directory.

docs/ignition/metadata.md
```diff
@@ -1,8 +1,6 @@
 ---
 title: Metadata
 weight: 10
-aliases:
-    - ../os/metadata.md
```
joaquimrocha: This was added in this PR, right? So in order to keep it cleaner, please remove it from the commit that introduces it (same for the other deletion).

joaquimrocha (changes requested):
Just one comment. Thanks.

docs/ignition/metadata.md
marga-kinvolk: I wanted to make it possible to understand what I had done: the first commit had the automatic changes and the commits after that had my manual fixes. But I guess it doesn't matter much how I got there.

I've now merged all the redirection-related commits into just one commit (including the fix to remove the trailing .md).

marga-kinvolk:
I've merged the commits as requested, can you please take another look? Thanks.

PR #146, opened 2020-12-31
docs/container-runtimes: Remove workaround for CRI plugin

pothos:
The latest Stable release does not need the workaround for Kubernetes
and containerd anymore.


vbatts (approved):
LGTM

PR #149, opened 2021-01-15
reference/developer-guides/sdk-modifying-flatcar: Host system requirements for arm64

pothos:
The arm64 target requires binary translation to be set up on the host system.
Document how to this can be set up, tested, and what the outcomes are.

### DIFF
--- a/docs/reference/developer-guides/sdk-modifying-flatcar.md
+++ b/docs/reference/developer-guides/sdk-modifying-flatcar.md
@@ -154,6 +154,7 @@ You can also rely on SSH key detection which normally works well when booting th
 #### Selecting the architecture to build
 
 `amd64-usr` and `arm64-usr` are the only targets supported by Flatcar.
+When setting up a new build target you can select it as default architecture, otherwise you need to specify the architecture for each step.
 
 ##### 64 bit AMD: The amd64-usr target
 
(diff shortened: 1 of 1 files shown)


docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -162,15 +163,40 @@ The `--board` option can be set to one of a few known target architectures, or s
 To create a root filesystem for the `amd64-usr` target beneath the directory `/build/amd64-usr/`:
 
 ```shell
-./setup_board --default --board=amd64-usr
+./setup_board [--default] --board=amd64-usr
 ```
 
+##### 64 bit ARM: The arm64-usr target
+
+The SDK assumes to run on an amd64 host system, relying on cross-compilation to create arm64 binaries.
```
marga-kinvolk, line 171: "assumes to run" is broken grammar, but I'm not sure what you're trying to say. Maybe "The SDK is expected to be run on ..."? Or maybe "SDK tools assume they are run on ..."

marga-kinvolk (approved):

docs/reference/developer-guides/sdk-modifying-flatcar.md
pothos: :smile: I think making it shorter would work: _The SDK runs on ..._

pothos: Changed it now

vbatts: i've meant to open an issue for how to bootstrap an arm64 host sdk...

PR #150, opened 2021-01-15
Fix link to notes-for-distributors

marga-kinvolk:
This was missed from my link-fix effort because it was hidden inside HTML code.


joaquimrocha (commented):
Not sure if it would work (though I think it does), but a more effective fix would be to add the missing base link to the end of https://raw.githubusercontent.com/kinvolk/flatcar-docs/main/docs/_index.md

marga-kinvolk:
I don't understand. This is fixing the path of a file. How does adding a link to the index change that?

joaquimrocha:
Because if you add the named link (as it's used before you changed it in this PR -- without any path in it), then it can be used everywhere without being changed.

PR #151, opened 2021-01-18
provisioning/terraform: Use new repo location

pothos:
The original repository was part of a talk presentation.
Now that we will add more examples, the repository was moved under
kinvolk/.


marga-kinvolk (approved):
Thanks

PR #152, opened 2021-01-21
Update the iscsi service name to reflect changes from update

sayanchowdhury:
# Update the iscsi service name to reflect changes from update

With the iscsi update, the systemd service that controls the
other services is `iscsi`

Signed-off-by: Sayan Chowdhury <sayan@kinvolk.io>


marga-kinvolk (approved):
LGTM

PR #153, opened 2021-02-03
container-runtimes: Update section about using containerd for K8s

pothos:
The default socket location was changed (in a backwards-compatible way
through a symlink on the old location) and the section about using the
upstream defaults was confusing because it also involved breaking
Docker which should be solved by disabling the Docker service and
socket, not through a broken configuration.


marga-kinvolk (approved):
LGTM.

Wow, this is so simple now :)

PR #154, opened 2021-02-09
installing/cloud/digitalocean: Update and extend instructions

pothos:
The instructions to run a droplet assumed a public image to be
    available which isn't.
    Update the instructions on how to create a custom image on the command
    line or in Terraform.

Tested in https://github.com/kinvolk/flatcar-terraform/pull/6

### DIFF
--- a/docs/installing/cloud/digitalocean.md
+++ b/docs/installing/cloud/digitalocean.md
@@ -11,8 +11,11 @@ On Digital Ocean, users can upload Flatcar Container Linux as a [custom image](h
 
 The _import URL_ should be `https://<channel>.release.flatcar-linux.net/amd64-usr/<version>/flatcar_production_digitalocean_image.bin.bz2`. See the [release page](https://www.flatcar-linux.org/releases/) for version and channel history.
 
-The following command will create a single droplet. For more details, check out [Launching via the API](#via-the-api).
+For more details, check out [Launching via the API](#via-the-api).
 
+At the end of the document there are instructions for deploying with Terraform.
+
+<!--
 <div id="do-images">
   <ul class="nav nav-tabs">
     <li><a class="active show" href="#stable" data-toggle="tab">Stable Channel</a></li>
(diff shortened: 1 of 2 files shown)


docs/installing/cloud/digitalocean.md
```diff
@@ -201,3 +192,157 @@ Now that you have a machine booted it is time to play around. Check out the [Fla
 
 [quick-start]: ../
 [docs]: ../../
+
+## Terraform
+
+The [`digitalocean`](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs) Terraform Provider allows to deploy machines in a declaratie way.
```
krnowak, line 198:
```suggestion
The [`digitalocean`](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs) Terraform Provider allows to deploy machines in a declarative way.
```

docs/installing/cloud/digitalocean.md
```diff
@@ -11,7 +11,9 @@ On Digital Ocean, users can upload Flatcar Container Linux as a [custom image](h
 
 The _import URL_ should be `https://<channel>.release.flatcar-linux.net/amd64-usr/<version>/flatcar_production_digitalocean_image.bin.bz2`. See the [release page](https://www.flatcar-linux.org/releases/) for version and channel history.
 
-The following command will create a single droplet. For more details, check out [Launching via the API](#via-the-api).
+For more details, check out [Launching via the API](#via-the-api).
+
+At the end of the document there are instructions for deploying with Terraform.
 
 <div id="do-images">
```
marga-kinvolk, line 18: This whole div adds nothing right now and it should be entirely removed. But I'd comment it out instead of deleting the code, so that when we have the images it's easier to re-add it.  I know we can always go back in git history, but it's more work and it's likely whoever is doing it might not remember or even know that this was there before.

docs/installing/cloud/digitalocean.md
```diff
@@ -201,3 +192,157 @@ Now that you have a machine booted it is time to play around. Check out the [Fla
...
+  }
+}
+
+resource "digitalocean_ssh_key" "first" {
+  name       = var.cluster_name
+  public_key = var.ssh_keys.0
+}
+
+resource "digitalocean_custom_image" "flatcar" {
+  name   = "flatcar-stable-2605.12.0"
```
marga-kinvolk, line 235: This will be outdated very quickly :(. Can't we use a version string instead?

Also, I find this embedding a rather complex terraform config in a single doc rather hard to read. Shouldn't there be a pointer to the examples repo somewhere along here?

pothos: Yes, added a variable for the version.

I see the Terraform code as something to copy from as needed but yes, will add a link.

docs/installing/cloud/digitalocean.md
pothos: Thanks

docs/installing/cloud/digitalocean.md
pothos: Commented out

marga-kinvolk (approved):
LGTM

PR #158, opened 2021-02-26
Fix kinvolk/Flatcar#351: Add headings for clarity

nathanhourt:
# Add headings to the static IP section to clarify the two alternative formats under consideration

The docs for static IP settings during Ignition run specify two alternative formats for the IP configuration passed on the kernel command line, but as written, this wasn't necessarily obvious on first reading. It appears as though the initial author had intended to insert subheadings but that they had gotten dropped. This commit adds headings, which clarify the meaning and make it easy to recognize the two formats and their differences on a first reading.

# How to use

Read the docs!

# Testing done

Prosaic alterations were tested using GitHub's markdown renderer, both before and after committing. HTML/CSS output from GitHub was subsequently processed using Firefox to render a final result. This result was transmitted to an IPS panel display. Emissions from the display were recorded using an organic photo-optical sensor array and transmitted in realtime to a neural network for processing and evaluation.


pothos (approved):

pothos:
Thanks :)

pothos:
(Will be live tomorrow)

PR #164, opened 2021-04-23
developer-guides: minor fixes

tormath1:
in this PR, we fix some wording / spelling issues - some light modifications have also been proposed.

### DIFF
--- a/docs/reference/developer-guides/sdk-modifying-flatcar.md
+++ b/docs/reference/developer-guides/sdk-modifying-flatcar.md
@@ -93,7 +93,7 @@ You may want to add the `PATH` export to your shell profile (e.g. `.bashrc`).
 If you want to build cork from sources instead of using an official release, you'll need golang installed as mantle is written in go.
 
 ```shell
-$ git clone git@github.com:flatcar-linux/mantle.git
+$ git clone git@github.com:kinvolk/mantle.git
 $ cd mantle
 $ ./build cork
 ```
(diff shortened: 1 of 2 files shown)


docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -253,6 +253,8 @@ You can also rebuild individual packages manually by running `emerge-<arch>-usr
 
 Now that we have all packages for the OS image either built or downloaded from the binary cache, we'll build a production base image:
 
+_Note_: The default login username is `core` and the password can be set with `./set_shared_user_password`
```
sayanchowdhury, line 256: Would be good to cross-link to the doc: https://kinvolk.io/docs/flatcar-container-linux/latest/reference/developer-guides/sdk-tips-and-tricks/#set-a-password-for-the-core-user-when-building-your-own-images

docs/reference/developer-guides/sdk-tips-and-tricks.md
```diff
@@ -147,18 +147,6 @@ $ cork enter
 $  ./set_version --dev-board --board-version amd64-usr/main-nightly --dev-sdk --sdk-version sdk-main-nightly
 ```
 
-## Caching git https passwords
```
sayanchowdhury, line 150: I don't use this, but this could be useful. Any reason to remove this?

docs/reference/developer-guides/sdk-modifying-flatcar.md
pothos: The `flatcar_production_qemu.sh` script uses ssh-agent to find the SSH keys and cloud platforms also have a way to populate the `.ssh` folder without an Ignition config. I wouldn't talk about setting a password because it can lead to situations where the deployed image is reachable on the Internet and a weak password was chosen.
There is also autologin on the serial console.

tormath1: > can lead to situations where the deployed image is reachable on the Internet and a weak password was chosen.

I agree. Should we remove any occurrence of password setting then ?

pothos: Still having it in https://kinvolk.io/docs/flatcar-container-linux/latest/reference/developer-guides/sdk-tips-and-tricks/#set-a-password-for-the-core-user-when-building-your-own-images is ok if it's needed for a certain scenario

sayanchowdhury: I'm okay with both as I treat the scope as a developer guide, and not for production. We had this in the document earlier, possibly removed it because we added autologin.

docs/reference/developer-guides/sdk-tips-and-tricks.md
tormath1: you're right. I misunderstood this section :P

sayanchowdhury:
I see that there are fixup , and revert commits. Are you planning to squash them? Also, I feel the commits can be merged into one or maybe 2.

tormath1:
@sayanchowdhury once PR approved, I'll rebase / squash the commits into two commits (one for the tips and tricks section and the other for the sdk section). :)

sayanchowdhury (approved):

PR #165, opened 2021-05-18
Fix broken links

kyontan:
# Fix broken links in docs

I found two broken links in the docs.
AFAIK, There's no other broken links (I tried to find same types of broken links using scripts)

- https://kinvolk.io/docs/flatcar-container-linux/latest/installing/bare-metal/booting-with-ipxe/#adding-a-custom-oem
- https://kinvolk.io/docs/flatcar-container-linux/latest/installing/bare-metal/booting-with-pxe/#installation

## How to use

Rebuilding pages may fix the problems.

## Testing done

I don't know how to build this, so I couldn't check whether this pr successfully fix the problems.


pothos (approved):
Thanks!

PR #168, opened 2021-07-01
flatcar/developer-guide: exit SDK before running script

tormath1:
since we don't ship qemu in the SDK with `slirp`, user
needs to use its own qemu to use user networking

Signed-off-by: Mathieu Tortuyaux <mathieu@kinvolk.io>

closes https://github.com/kinvolk/Flatcar/issues/420

### DIFF
--- a/docs/reference/developer-guides/sdk-modifying-flatcar.md
+++ b/docs/reference/developer-guides/sdk-modifying-flatcar.md
@@ -269,9 +269,9 @@ In general, `image_to_vm.sh` will read the generic disk image, install any vendo
 
 ### Booting
 
-`image_to_vm.sh` for the qemu format the will also generate a wrapper script to launch a Flatcar VM with qemu. You can boot the VM with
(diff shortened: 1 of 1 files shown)


docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -269,9 +269,9 @@ In general, `image_to_vm.sh` will read the generic disk image, install any vendo
 
 ### Booting
 
-`image_to_vm.sh` for the qemu format the will also generate a wrapper script to launch a Flatcar VM with qemu. You can boot the VM with
+`image_to_vm.sh` will also generate a wrapper script to launch a Flatcar VM with qemu. After exiting the SDK, you can boot the VM with:
```
pothos, line 272:
```suggestion
`image_to_vm.sh` will also generate a wrapper script to launch a Flatcar VM with qemu. In a new terminal, without entering the SDK, you can boot the VM with:
```

tormath1:
@pothos thanks for review. Suggestion has been addressed :)

pothos (approved):

PR #170, opened 2021-07-26
docs: Add note for SDK directories mounted with `nodev`

christarazi:
This was causing trouble for me when I was trying to use `cork` inside
`/tmp`.

Suggested-by: Jeremi Piotrowski <jpiotrowski@microsoft.com>
Suggested-by: Thilo Fromm <thilo@kinvolk.io>
Signed-off-by: Chris Tarazi <tarazichris@gmail.com>


christarazi:
@jepio @t-lo Thanks for the help, here's the PR to update the docs regarding the "permission denied" errors with `cork` inside `/tmp`.

jepio:
Thanks! Hope this saves someone else from that trap 😄

PR #172, opened 2021-07-30
cgroup v2 docs

jepio:
Add a page under `container runtimes` that explains the details of the cgroupv2 migration and impact.

### DIFF
--- a/docs/container-runtimes/switching-to-unified-cgroups.md
+++ b/docs/container-runtimes/switching-to-unified-cgroups.md
@@ -8,18 +8,32 @@ aliases:
 
 With the upgrade to Systemd v248, Flatcar Linux has migrated to the unified
 cgroup hierarchy (aka cgroup v2). Much of the container ecosystem has already
-moved to default to cgroup v2. Cgroup v2 also brings new and exciting features related to
-eBPF tracing.
+moved to default to cgroup v2. Cgroup v2 brings new and exciting features in
+areas such as eBPF and rootless containers.
 
 Flatcar nodes deployed prior to this change will be kept on cgroup v1 (legacy
 hierarchy) and will require manual migration. During an update from an older
-Flatcar version, a post installation script adds the kernel command line
-parameter `systemd.unified_cgroup_hierarchy=0` to `/usr/share/oem/grub.cfg`. To
-migrate such nodes to cgroup v2, either remove the line or change it to
-`systemd.unified_cgroup_hierarchy=1`.
+Flatcar version, a post update script does two things:
 
-Newly deployed nodes will default to cgroup v2.  To revert to cgroup v1 on such
-nodes, use the following Ignition snippet (here as CLC YAML to be transpiled to Ignition JSON):
+* adds the kernel command line parameters `systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller`
+  to `/usr/share/oem/grub.cfg`
+* creates a systemd drop-in unit at `/etc/systemd/system/containerd.service.d/10-use-cgroupfs.conf` that
+  configures `containerd` to use systemd for cgroups.
+
+# Migrating old nodes to unified cgroups
+
+To undo the changes performed by the post update script, execute the following commands as root (or using `sudo`):
+
+```bash
+rm /etc/system/system/containerd.service.d/10-use-cgroupfs.conf
+sed -i -e '/systemd.unified_cgroup_hierarchy=0/d' /usr/share/oem/grub.cfg
+reboot
+```
+
+# Starting new nodes with legacy cgroups
+
+Nodes deployed with the release incorporating the described changes use cgroups v2 by default. To revert to cgroups v1 on new
+nodes during provisioning, use the following Ignition snippet (here as CLC YAML to be transpiled to Ignition JSON):
 
 ```yaml
 storage:
@@ -35,22 +49,32 @@ storage:
       append: true
       contents:
         inline: |
-          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0"
+          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller"
+systemd:
+  units:
+    - name: containerd.service
+      dropins:
+        - name: 10-use-cgroupfs.conf
+          contents: |
+            [Service]
+            Environment=CONTAINERD_CONFIG=/usr/share/containerd/config-cgroupfs.toml
 ```
 
-However, this setting doesn't take effect on the first boot, and a reboot is required before the snippet becomes active.
+However, the kernel commandline setting doesn't take effect on the first boot, and a reboot is required before the snippet becomes active.
 
 Beware that over time it is expected that upstream projects will drop support for cgroup v1.
 
+# Kubernetes
+
 The unified cgroup hierarchy is supported starting with Docker v20.10 and
 Kubernetes v1.19. Users that need to run older version will need to revert to
-cgroup v1, but are urged to find a migration path. Flatcar now ships with Docker
+cgroups v1, but are urged to find a migration path. Flatcar now ships with Docker
 v20.10, older versions can be deployed following the instructions on [running custom docker versions](use-a-custom-docker-or-containerd-version).
 
 Flatcar nodes that had Kubernetes deployed on them before the introduction of
-cgroup v2 should be careful when migrating. Depending on the deployment method,
+cgroups v2 should be careful when migrating. Depending on the deployment method,
 the `cgroupfs` cgroup driver may be hardcoded in the `kubelet` configuration.
-Cgroup v2 is only supported with the `systemd` cgroup driver. See [configuring a cgroup driver][kube-cgroup-docs] in the Kubernetes documentation for a discussion of cgroup drivers and how to migrate nodes. We recommend redeploying Kubernetes on fresh nodes instead of migrating inplace.
+Cgroups v2 are only supported with the `systemd` cgroup driver. See [configuring a cgroup driver][kube-cgroup-docs] in the Kubernetes documentation for a discussion of cgroup drivers and how to migrate nodes. We recommend redeploying Kubernetes on fresh nodes instead of migrating inplace.
 
 The cgroup driver used by `kubelet` should be the same as the one used by `docker` daemon. `docker` defaults to `systemd` cgroup driver when started on a system running cgroup v2 and `cgroupfs` when running on a system with cgroup v1. The cgroup driver can be explicitly configured for `docker` by either creating/extending `/etc/docker/daemon.json`:
 ```json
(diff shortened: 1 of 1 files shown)


pothos:
Looks good, maybe we also cover containerd (https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd) and cri-o (https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-driver) but I haven't tried it yet to verify

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,83 @@
...
+
+Flatcar nodes deployed prior to this change will be kept on cgroup v1 (legacy
+hierarchy) and will require manual migration. During an update from an older
+Flatcar version, a post installation script adds the kernel command line
+parameter `systemd.unified_cgroup_hierarchy=0` to `/usr/share/oem/grub.cfg`. To
+migrate such nodes to cgroup v2, either remove the line or change it to
+`systemd.unified_cgroup_hierarchy=1`.
+
+Newly deployed nodes will default to cgroup v2.  To revert to cgroup v1 on such
+nodes, use the following ignition snippet:
```
pothos, line 22:
```suggestion
nodes, use the following Ignition snippet (here as CLC YAML to be transpiled to Ignition JSON):
```

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,83 @@
...
+    - filesystem: "OEM"
+      path: "/grub.cfg"
+      mode: 0644
+      append: true
+      contents:
+        inline: |
+          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0"
+```
+
+A reboot is required before the snippet becomes active.
```
pothos, line 41:
```suggestion
However, this setting doesn't take effect on the first boot, and a reboot is required before the snippet becomes active.
```

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,83 @@
...
+# Kubernetes Container Runtimes
+
+When deploying Kubernetes through `kubeadm`, the default container runtime on Flatcar will be `dockershim`. In this setup, `kubelet` talks to `dockershim`, which talks to `docker`, which interfaces with `containerd`. The `SystemdCgroup` setting in `containerd`'s `config.toml` is ignored. `docker`'s cgroup driver and `kubelet` cgroup driver settings must match. Starting with Kubernetes v1.22, `kubeadm` will default to the `systemd` `cgroupDriver` setting if no setting is provided explicitly. Out of the box, Flatcar defaults are compatible with Docker and Kubernetes defaults - everything will use `systemd` cgroup driver.
+
+If users choose the `containerd` runtime, they must ensure that `containerd`'s support for `SystemdCgroups` is enabled. To do that, follow the instructions on
+[how to customize containerd configuration](customizing-docker) and add the following to `config.toml`:
+```toml
+[plugins.cri.containerd.runtimes.runc]
+  [plugins.cri.containerd.runtimes.runc.options]
+    SystemdCgroup = true
```
pothos, line 76: Can we add this as default in Flatcar's containerd config?

jepio: I thought about it - but changing this breaks kubernetes clusters because a cgroupfs-kubelet will not tolerate this value being SystemdCgroup. The one thing we could do is make this work like `lokomotive` here https://github.com/kinvolk/lokomotive/blob/master/assets/terraform-modules/node/templates/node.yaml.tmpl#L118 - have a script run before containerd starts that generates an config snippet that is included. What do you think?

pothos: If I understand it correctly, the default is cgroupfs and won't work for new deployments? That's not really ideal… Yes, writing out the right choice for the runtime-generated containerd config file is an option.

The other is that we can add an entry here, too, but still people would need to find it:
https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-from-docker-to-containerd-for-kubernetes/

pothos: Maybe I didn't get the part of the default value, in case containerd selects systemd as default after a cgroupv2 environment was found, then we should be fine without explicitly adding an entry, but if that it's the case we would need to have a script for that in torcx.

pothos: Ok, I think this here is fine since there is no autodiscovery but we need to call it out in https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-from-docker-to-containerd-for-kubernetes/, too.

I would also like to have a date where the default gets switched to `true`, and have this stated in the beginning of this doc (and possibly in the motd).

jepio: When docker is the CRI, the containerd SystemdCgroup setting has no effect. Docker does autodetection, containerd doesn't.  I was originally thinking we can get away with keeping the containerd setting in at the default, and changing it 2 releases (1 month) later, but now i'm not so sure. I fear that adding autodetection increases the complexity. I need to sleep on it to have a better idea.
For sure i'll update the docs you referenced.

pothos: Yes, autodetection is not the best idea, and it's also not easy because it has to consider whether cgroupv1 or cgroupv2 is used and if cgroupv1 is used, whether the kubelet wants to use `cgroupfs` or `systemd`…

As said, the compromise I see is to keep the default to `cgroupfs` to not break existing systems but also communicate that the default will be switched.
Another option is yet antoher postinstall hook where we could freeze the containerd config (in case people use the one from torcx) to `cgroupfs` and do the switch to systemd now.

pothos: The decision was to extend the postinstall hook to also freeze the containerd config: https://github.com/kinvolk/update_engine/pull/13

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Switching to Unified Cgroups
+linktitle: Switching to unified cgroups
+description: Overview of changes necessary to use unified cgroups with Kubernetes
+weight: 20
+aliases:
+---
+
+With the upgrade to Systemd v248, Flatcar Linux has migrated to the unified
```
sayanchowdhury, line 9: We might not migrate to systemd v248, rather we would directly jump to v249. Do you think if it will confuse people if we mentioned that since `systemd v248, Flatcar Linux has migrated to the unified cgroup hierachy`?

jepio: I didn't know how to express this better - do you think i should just put a provisional release number in (2970?) and then correct if the release happens a day +/-?

sayanchowdhury: Yes, I think using a provisional release number would be better compared to systemd version.

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,120 @@
...
+areas such as eBPF and rootless containers.
+
+Flatcar nodes deployed prior to this change will be kept on cgroup v1 (legacy
+hierarchy) and will require manual migration. During an update from an older
+Flatcar version, a post update script does two things:
+
+* adds the kernel command line parameters `systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller`
+  to `/usr/share/oem/grub.cfg`
+* creates a systemd drop-in unit at `/etc/systemd/system/containerd.service.d/10-use-cgroupfs.conf` that
+  configures `containerd` to use systemd for cgroups.
```
pothos:
```suggestion
  configures `containerd` to use `cgroupfs` for cgroups.
```

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Migrating old nodes to unified cgroups
+
+To undo the changes performed by the post update script, execute the following commands as root (or using `sudo`):
+
+```bash
+rm /etc/system/system/containerd.service.d/10-use-cgroupfs.conf
+sed -i -e '/systemd.unified_cgroup_hierarchy=0/d' /usr/share/oem/grub.cfg
```
pothos, line 29:
```suggestion
sed -i -e '/systemd.unified_cgroup_hierarchy=0/d' /usr/share/oem/grub.cfg
sed -i -e '/systemd.legacy_systemd_cgroup_controller/d' /usr/share/oem/grub.cfg
```

docs/container-runtimes/switching-to-unified-cgroups.md
jepio: Thanks, found that one just a second ago.

docs/container-runtimes/switching-to-unified-cgroups.md
jepio: Done

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,120 @@
...
+---
+kind: KubeletConfiguration
+apiVersion: kubelet.config.k8s.io/v1beta1
+cgroupDriver: systemd
+```
+
+## Containerd
+
+If users choose the `containerd` runtime, they must ensure that `containerd`'s setting for `SystemdCgroup` is consistent with `kubelet` and `docker` settings. Flatcar enables `SystemdCgroup` by default for `containerd`. Users may change the setting to suit their deployment. To do that, follow the instructions on
+[how to customize containerd configuration](customizing-docker) and add the relevant lines to your `config.toml`:
```
pothos, line 108:
```suggestion
If users choose the `containerd` runtime, they must ensure that `containerd`'s setting for `SystemdCgroup` is consistent with `kubelet` and `docker` settings. Flatcar enables `SystemdCgroup` by default for `containerd`. Users may change the setting to suit their deployment.
If you maintain your own containerd configuration or did follow the instructions on
[how to customize containerd configuration](customizing-docker), you should add the relevant lines to your `config.toml`:
```

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -0,0 +1,121 @@
...
+  to `/usr/share/oem/grub.cfg`
+* creates a systemd drop-in unit at `/etc/systemd/system/containerd.service.d/10-use-cgroupfs.conf` that
+  configures `containerd` to keep using cgroupfs for cgroups.
+
+# Migrating old nodes to unified cgroups
+
+To undo the changes performed by the post update script, execute the following commands as root (or using `sudo`):
+
+```bash
+rm /etc/system/system/containerd.service.d/10-use-cgroupfs.conf
```
jepio, line 28:
```suggestion
rm /etc/systemd/system/containerd.service.d/10-use-cgroupfs.conf
```

pothos (approved):

PR #178, opened 2021-09-02
Update verify-images.md

jepio:
Fix the link to the signing-key on the website, and update the text to match currently used keys.


pothos (approved):
Thanks

I also have some more changes in https://github.com/kinvolk/flatcar-docs/pull/179 but can rebase

PR #179, opened 2021-09-02
Use correct image signing key URL

pothos:
The old website has redirects in place for many parts but not for all,
and thus, the image signing key was outdated.
Use the right URL which users will also find on the website.


jepio:
Can you pull in the gpg command output from https://github.com/kinvolk/flatcar-docs/pull/178

jepio (approved):

PR #180, opened 2021-09-04
Update installing-to-disk.md

ArkaprabhaChakraborty:
#88 Document dependencies of flatcar-install script 

Added names of the necessary binaries required by the flatcar-install script.

### DIFF
--- a/docs/installing/bare-metal/installing-to-disk.md
+++ b/docs/installing/bare-metal/installing-to-disk.md
@@ -9,6 +9,20 @@ aliases:
     - ../../os/installing-to-disk
     - ../../bare-metal/installing-to-disk
 ---
+## Required Binaries
+The following Binaries are required to run the flatcar-install script:
+```
+lbzip2 or bzip2 
+wget
+coreutils (gnu core utils package)
+udev (for alpine images it's eudev)
+```
+Some optional binaries are:
+```
+gnupg2
+gawk
+```
+
 
 ## Install script
 
--- a/docs/installing/bare-metal/installing-to-disk.md
+++ b/docs/installing/bare-metal/installing-to-disk.md
@@ -17,7 +17,7 @@ lbzip2 or bzip2
 mount, lsblk  (often found in the util-linux packaage)
 wget
 grep
-cp, dd, mkfifo, mkdir, rm, tee (found in coreutils or gnu core utils package)
+cp, dd, mkfifo, mkdir, rm, tee (often found in the GNU coreutils package or as part of busybox)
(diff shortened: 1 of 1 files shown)


docs/installing/bare-metal/installing-to-disk.md
```diff
@@ -9,6 +9,20 @@ aliases:
...
+lbzip2 or bzip2 
+wget
+coreutils (gnu core utils package)
+udev (for alpine images it's eudev)
+```
+Some optional binaries are:
+```
+gnupg2
+gawk
+```
```
pothos, line 24:
```suggestion
## Required dependencies
If you want to use the `flatcar-install` script on some other environment than Flatcar Container Linux, ensure that the following binaries are present:
```
bash
lbzip2 or bzip2 
wget
mount, lsblk (often found in the util-linux package)
grep
cp, dd, mkfifo, mkdir, rm, tee (often found in the GNU coreutils package)
udevadm (found in the udev package, or for Alpine, in eudev)
gpg, gpg2 (found in gnupg2)
gawk (often found in the GNU gawk package)
```

Comment (pothos) [reply docs/installing/bare-metal/installing-to-disk.md:24]: I can't have the correct code block start and end sequence in the suggestion itself, the `// ` was just added as a workaround and should not be committed.

Comment (ArkaprabhaChakraborty) [reply docs/installing/bare-metal/installing-to-disk.md:24]: Okay Ill make the changes

File: docs/installing/bare-metal/installing-to-disk.md [hunk 2]
```diff
@@ -9,6 +9,20 @@ aliases:
...
 ---
+## Required Dependencies
+If you want to use the `flatcar-install` script on some other environment than Flatcar Container Linux, ensure that the following binaries are present:
+```
+bash
+lbzip2 or bzip2 
+mount, lsblk  (often found in the util-linux packaage)
+wget
+grep
+cp, dd, mkfifo, mkdir, rm, tee (found in coreutils or gnu core utils package)
```
pothos, line 20:
```suggestion
cp, dd, mkfifo, mkdir, rm, tee (often found in the GNU coreutils package or as part of busybox)
```
I would write it like that - the coreutils package is normally from the GNU coreutils project, but on Alpine these would be provided as part of busybox

ArkaprabhaChakraborty: Okay Ill change that

ArkaprabhaChakraborty: yes for alpine its provided as part of busy box

pothos (approved):
Thanks!

PR #183, opened 2021-09-15
provisioning/ignition/boot-process: document how to simulate first boot

pothos:
Running Ignition on an already booted root filesystem has many traps.
One of them is that systemd service presets are not reevaluated,
resulting in all new supposed-to-be-enabled services being disabled by
default.

Document how to invalidate the machine-id file so that systemd
reevaluates the service presets.

## How to use/Testing done

Booted a qemu image, then created the `first_boot` flag file and booted it again with an Ignition config set. The service in the Ignition config does not get enabled unless one also did invalidate the machine-id file.


t-lo (approved):
LGTM - thanks for catching this, and for the swift doc update!

jepio:
What do you think about removing the /etc/machine-id file instead? The method you describe feels like an obscure implementation detail whereas removing feels more intuitive to outsiders.

pothos:
Yes, removing is probably the better idea (https://manpages.debian.org/testing/systemd/machine-id.5.en.html#FIRST_BOOT_SEMANTICS)

PR #184, opened 2021-11-08
developer-guides: Update urls/cork/releases used in SDK example

jepio:
# developer-guides: Update urls/cork/releases used in SDK example

Multiple changes happened in October 2021:
* the new default URL in cork (mirror.release.flatcar-linux.net/sdk) only has releases that happened after
  October 2021, so update the SDK versions used in examples to recent ones.
* the website is moving from kinvolk.io to flatcar-linux.org so update url to releases.json (old url is not a
  redirect and does not get updated) and to GPG key (old url works). The GPG output also needed updating.
* cork 0.17.0 has updated urls so reference that.

## How to use

Follow the instructions.

### DIFF
--- a/docs/reference/developer-guides/sdk-modifying-flatcar.md
+++ b/docs/reference/developer-guides/sdk-modifying-flatcar.md
@@ -59,9 +59,9 @@ After the installation finished, `cork enter` may be used to enter the SDK chroo
 First, download the cork utility and verify it with the signature:
 
 ```shell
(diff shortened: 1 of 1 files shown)


t-lo (approved):
Great work Jeremi, thank you! Particularly like the thoroughness of updating the channel versions in the examples.

jepio:
Thanks, the channel versions are especially important to update because the old ones are not available on the mirror server.

PR #187, opened 2021-11-23
add `gawk` to the requirements for hetzner

syphernl:
# [Title: describe the change in one sentence]

The Hetzner Rescue System runs on Debian 11 but seems to be missing `gawk` which is required by the flatcar installer.

## How to use

Follow the instructions

## Testing done

* Tested the config as-is in a clean Hetzner Rescue system, which showed an error about missing `gawk`
* Then added `gawk` and the process worked fine

### DIFF
--- a/docs/installing/cloud/hetzner.md
+++ b/docs/installing/cloud/hetzner.md
@@ -23,6 +23,7 @@ Select any OS like Debian when you create the instance but boot into the `linux6
 Connect via SSH and download and run the `flatcar-install` script:
 
 ```sh
+apt -y install gawk
 curl -fsSLO --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 https://raw.githubusercontent.com/kinvolk/init/flatcar-master/bin/flatcar-install
 chmod +x flatcar-install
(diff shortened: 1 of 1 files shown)


docs/installing/cloud/hetzner.md
```diff
@@ -23,6 +23,7 @@ Select any OS like Debian when you create the instance but boot into the `linux6
 Connect via SSH and download and run the `flatcar-install` script:
 
 ```sh
+apt -y install gawk
```
pothos, line 26:
```suggestion
apt update
apt -y install gawk
```
Thanks, in general a repo update is also needed to ensure that the binary package exists

syphernl: I have installed it fine without an explicit `apt update`. But as this is no guarantee for the future it might be wise to add indeed.

pothos (approved):

docs/installing/cloud/hetzner.md
pothos: There is also https://github.com/flatcar-linux/flatcar-terraform/tree/main/hetzner which has the same steps

PR #188, opened 2021-11-24
azure: add more info on how to use Flatcar

pothos:
The Marketplace image has some quirks and the terminology and functions
are a bit difficult to understand when provisioning Flatcar instances.

Add more info on how to use the Marketplace image, what its plan values
are, and how instance userdata and user provisioning works.

## How to use

Read https://www.flatcar-linux.org/docs/latest/installing/cloud/azure/ and see if the additions make sense.

### DIFF
--- a/docs/installing/cloud/azure.md
+++ b/docs/installing/cloud/azure.md
@@ -114,11 +114,28 @@ $ az vm create --name node-1 --resource-group group-1 --admin-username core --cu
   </div>
 </div>
 
+You can use both image offers `flatcar-container-linux` and `flatcar-container-linux-free`, the contents are the same.
+The SKU, which is the third element of the image URN, relates to one of the release channels and also depends on whether to use HyperV Generation 1 or 2.
+Generation 1 instance types use the channel names `alpha`, `beta` or `stable` as is; for Generation 2 instance types please append `-gen2` to the channel name, i.e., use one of `alpha-gen2`, `beta-gen2` or `stable-gen2`.
+This means the Gen 2 image URN for the above example for a Stable release becomes `flatcar-container-linux:stable-gen2:2345.3.0`.
+
+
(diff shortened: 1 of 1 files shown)


pothos:
Maybe it's good to check if ` --plan-name` and the others are needed for VM creation, too, where they are missing at the moment.

docs/installing/cloud/azure.md
```diff
@@ -114,12 +114,25 @@ $ az vm create --name node-1 --resource-group group-1 --admin-username core --cu
   </div>
 </div>
 
+You can use both image offers `flatcar-container-linux` and `flatcar-container-linux-free`, and depending on your instance type the `CHANNEL-gen2` images for HyperV Generation 2 images.
```
t-lo, line 117:
```suggestion
You can use both image offers `flatcar-container-linux` and `flatcar-container-linux-free`. The release channels  `alpha`, `beta` or `stable` can be selected via Azure's `--plan` option. Generation 1 instance types use the plan names as is; for Generation 2 instance types please append `-gen2` to the plan name (i.e. use one of `alpha-gen2`, `beta-gen2` or `stable-gen2`).
```

t-lo (approved):
LGTM, one suggestion added to be more explicit about the channel / plan relation.

docs/installing/cloud/azure.md
pothos: In the example above `--image kinvolk:flatcar-container-linux:stable:2345.3.0` is used, doesn't this mean gen 2 would be `--image kinvolk:flatcar-container-linux:stable-gen2:2345.3.0` instead of an additional `--plan` argument?

jepio: @pothos is correct. In the URN syntax, "(alpha|beta|stable)(-gen2)?" is actually the "sku". The "plan" argument is just metadata.

PR #189, opened 2021-11-25
Add example for whole disk as swap

cpswan:
# Add example for whole disk as swap

Container Linux config sample with sections for disks, filesystems and the systemd unit to mount swap.

## How to use

Create an ignition file including the transpiled content from the example.

Start a VM with an additional disk that will be mounted as /dev/sdb on first boot.

## Testing done

This has been tested on GCE using this script to start VMs:

```bash
#!/bin/bash
VMNAME=test-123
ZONE=us-central1-a
PROJECT=myproject
gcloud compute instances create "$VMNAME" --project="$PROJECT" \
--zone "$ZONE" --machine-type e2-standard-2 \
--network-interface=network-tier=PREMIUM,subnet="$PROJECT"vpc-01 \
--create-disk=auto-delete=yes,boot=yes,device-name="$VMNAME"-boot,image-project=kinvolk-public,image-family=flatcar-stable,mode=rw,size=100,type=projects/"$PROJECT"/zones/"$ZONE"/diskTypes/pd-standard \
--create-disk=device-name="$VMNAME"-swap,mode=rw,name="$VMNAME"-swap,size=32,type=projects/"$PROJECT"/zones/"$ZONE"/diskTypes/pd-ssd \
--metadata-from-file user-data=config_with_swap.ign 
```

### DIFF
--- a/docs/setup/storage/adding-swap.md
+++ b/docs/setup/storage/adding-swap.md
@@ -126,3 +126,41 @@ systemd:
         ExecStart=/usr/sbin/mkswap /var/vm/swapfile1
         RemainAfterExit=true
 ```
+
+## Using a dedicated swap disk
+
+The following Container Linux config sets up `/dev/sdb` to be used as swap:
+
+```yaml
+storage:
+  disks: 
+    - device: /dev/sdb 
+      wipe_table: true 
+      partitions: 
+        - label: swap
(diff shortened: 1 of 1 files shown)


jepio (approved):
Awesome, thanks for contributing this!

PR #190, opened 2021-11-26
update-strategies: document how to do manual updates

pothos:
The existing way of disabling automatic updates with masking the
services was difficult when the user still wanted to do manual updates
because one needs to know that unmasking needs to be kept for the first
minutes after the reboot.

Document an easier and safer way that uses the update-engine config
file to disable automatic updates.

## Testing done

Tested the CLC snippet and that update-engine correctly does not find updates but is still running to be able to mark the partition as successful


t-lo:
Can't we keep both options in the docs and call out it's preferred to specify an invalid server since that makes it easier to run one-shot updates?

pothos:
Sure, added it as alternative

t-lo (approved):
LGTM, thank you Kai!

PR #191, opened 2021-12-01
Rework update and rollback docs

pothos:
- Improvements for the migration script
- Introduce the flatcar-update script from https://github.com/flatcar-linux/init/pull/53
- Reworked to have simpler instructions to follow and added some hints for things to check

### DIFF
--- a/docs/migrating-from-coreos/update-from-container-linux.md
+++ b/docs/migrating-from-coreos/update-from-container-linux.md
@@ -10,74 +10,10 @@ If you already have CoreOS Container Linux clusters and can't or don't want to f
 
 **NOTE:** General differences when [migrating from CoreOS Container Linux][migrate-from-container-linux] also apply.
 
-At [the end of the section](#all-steps-in-one-script) you can find the [update-to-flatcar.sh](https://raw.githubusercontent.com/kinvolk/flatcar-docs/main/update-to-flatcar.sh) script that does all steps for you.
 
-## Getting the public update key
+## The migration script
 
-First, you need to get Flatcar's public update key:
-
-```shell
-curl -L -o /tmp/key https://raw.githubusercontent.com/flatcar-linux/coreos-overlay/flatcar-master/coreos-base/coreos-au-key/files/official-v2.pub.pem
-```
-
-Since the `/usr` partition is read-only, to allow the updater to use the new key, you need to bind-mount it:
-
-```shell
-sudo mount --bind /tmp/key /usr/share/update_engine/update-payload-key.pub.pem
-```
-
-## Modifying the configuration files
-
-Migrate your `/etc/coreos/update.conf` settings but keep a symlink:
-
-```shell
-sudo mv /etc/coreos /etc/flatcar
-sudo ln -s flatcar /etc/coreos
-```
-
-Now, you need to point `update_engine` to Flatcar's update server by setting the `SERVER` configuration option in `/etc/flatcar/update.conf`:
-
-```shell
-SERVER=https://public.update.flatcar-linux.net/v1/update/
-```
-
-To make sure you get an update even if you're running the same CoreOS Container Linux version as the latest Flatcar Container Linux, you need to force an update by clearing the current version number from the `release` file.
-This file also lives in the `/usr` partition so you need to do a bind-mount again:
-
-```shell
-cp /usr/share/coreos/release /tmp
-sudo mount --bind /tmp/release /usr/share/coreos/release
-```
-
-Then, you need to edit `/usr/share/coreos/release` and replace the value of `COREOS_RELEASE_VERSION` with `0.0.0`:
-
-```shell
-COREOS_RELEASE_VERSION=0.0.0
-```
-
-Migrate any cloud-config `user_data` if it exists in `/var/lib/coreos-install/user_data` (e.g., in bare metal installations):
-
-```shell
-test -d /var/lib/coreos-install && sudo ln -sn /var/lib/coreos-install /var/lib/flatcar-install
-```
-
-## Restart service and reboot
-
-After that, restart the update service so it rescans the edited configuration. Initiate an immediate update.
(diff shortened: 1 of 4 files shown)


docs/migrating-from-coreos/update-from-container-linux.md
```diff
@@ -90,6 +26,16 @@ Done, please reboot now
 core@host ~ $ sudo systemctl reboot
 ```
 
+If it fails due to SSL connection issues from outdated certificates, you can also download the update payload of the latest Stable release through plain HTTP and use the `flatcar-update` script instead:
+
+```shell
+$ VER=$(curl -fsSL https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt | grep FLATCAR_VERSION= | cut -d = -f 2)
```
tormath1, line 32: > If it fails due to SSL connection issues from outdated certificates

Maybe we should run `curl` in insecure mode too, otherwise it will fail to verify `stable.release.flatcar-linux.net` certificate?

```suggestion
$ VER=$(curl -fsSL --insecure https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt | grep FLATCAR_VERSION= | cut -d = -f 2)
```

docs/setup/debug/manual-rollbacks.md
```diff
@@ -14,6 +14,28 @@ This section describes the automated upgrade process, performing a manual rollba
 
 **Note:** Neither performing a manual rollback nor forcing a channel downgrade are recommended.
 
+## Automated rollbacks
+
+The rollback to the previously installed version is done by GRUB and happens automatically if `update-engine` had no chance to mark the version as successful.
+This marking happens when the new version is booted and keeps running for around two minues, at which point `update-engine` will mark the version as successful (how this works in detail is explained below).
```
tormath1, line 20:
```suggestion
This marking happens when the new version is booted and keeps running for around two minutes, at which point `update-engine` will mark the version as successful (how this works in detail is explained below).
```

docs/setup/debug/manual-rollbacks.md
```diff
@@ -14,6 +14,28 @@ This section describes the automated upgrade process, performing a manual rollba
...
+
+To extend the automatic rollback logic to cover your important systemd services, you could make them a requirement for the `update-engine.service`.
```
tormath1, line 22:
```suggestion
To extend the automatic rollback logic to cover your important systemd services, you could make them as requirement for the `update-engine.service`.
```

docs/setup/debug/manual-rollbacks.md
```diff
@@ -14,6 +14,28 @@ This section describes the automated upgrade process, performing a manual rollba
...
+
+Note that `update-engine` will still try to update which can cause a loop with disruptions due to the reboots.
+You can disable automatic updates by setting `SERVER=disabled` in `/etc/flatcar/update.conf`.
+
+## Rollback with `flatcar-update`
+
+While you can roll back to the previously installed version manually with the rest of this guide, you can also install any version to the inactive partition with the `flatcar-update` tool.
```
tormath1, line 29:
```suggestion
While you can rollback to the previously installed version manually with the rest of this guide, you can also install any version to the inactive partition with the `flatcar-update` tool.
```

To stay consistent with other `rollback`.

docs/migrating-from-coreos/update-from-container-linux.md
pothos: Good catch, it's probably better to use http instead under this assumption

tormath1: Yes - I was assuming a redirection from `http` to `https`. Using `http` will be enough then. :)

docs/migrating-from-coreos/update-from-container-linux.md
```diff
@@ -90,6 +26,16 @@ Done, please reboot now
 core@host ~ $ sudo systemctl reboot
 ```
 
+If it fails due to SSL connection issues from outdated certificates, you can also download the update payload of the latest Stable release through plain HTTP and use the `flatcar-update` script instead:
+
+```shell
+$ VER=$(curl -fsSL http://stable.release.flatcar-linux.net/amd64-usr/current/version.txt | grep FLATCAR_VERSION= | cut -d = -f 2)
+$ wget "http://update.release.flatcar-linux.net/amd64-usr/$VER/flatcar_production_update.gz"
+$ wget https://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/bin/flatcar-update
```
tormath1, line 34: Same thing with `https` here - and I guess this time we have a redirect from `github.com`.
```suggestion
$ wget --no-check-certificate https://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/bin/flatcar-update
```

With a warning to double check the `flatcar-update` (like you did elsewhere)

pothos: Ah, yes, `http` works here, too. I remember that sometimes the curl/wget would still refuse to download and `--insecure` wasn't enough for curl, I would simply use http.

tormath1: Ok but to me `github.com` will force the redirect to `https`.
Exemple:
```
$ curl --head http://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/LICENSE
HTTP/1.1 301 Moved Permanently
Server: Varnish
Retry-After: 0
Location: https://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/LICENSE
$ wget http://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/LICENSE
URL transformed to HTTPS due to an HSTS policy
--2021-12-02 12:34:37--  https://raw.githubusercontent.com/flatcar-linux/init/flatcar-master/LICENSE
```

pothos: Oh, switched to using both http and the insecure options in case of redirects

tormath1 (approved):

PR #192, opened 2021-12-06
Document OEM update hook and improve docs on reboot managers

pothos:
The update post-install hook can be useful to run custom steps. Since it's not documented yet we do so here. The reboot window docs got moved closer to the reboot strategy docs.

### DIFF
--- a/docs/setup/releases/update-strategies.md
+++ b/docs/setup/releases/update-strategies.md
@@ -9,9 +9,17 @@ aliases:
 
 The overarching goal of Flatcar Container Linux is to secure the Internet's backend infrastructure. We believe that automatically updating the operating system is one of the best tools to achieve this goal.
 
-We realize that each Flatcar Container Linux cluster has a unique tolerance for risk and the operational needs of your applications are complex. In order to meet everyone's needs, there are three update strategies that we have developed based on feedback during our alpha period.
+We realize that each Flatcar Container Linux cluster has a unique tolerance for risk and the operational needs of your applications are complex. In order to meet everyone's needs, there are different update/reboot strategies that we have developed.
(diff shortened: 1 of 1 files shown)


pothos:
Also added FLUO and kured references. Made the docs call out locksmithd as the default reboot manager and documented a bit more what it does by default and how to configure it.

docs/setup/releases/update-strategies.md
```diff
@@ -160,22 +215,11 @@ update_engine_client -reset_status
...
-  window_length: 1h
-```
-
-This will configure a Flatcar Container Linux machine to follow the `reboot` strategy, and thus when an update is ready it will simply reboot instead of attempting to grab a lock in etcd. This machine however has also been configured to only reboot between 04:00 and 05:00 on Thursdays, so if an update occurs outside of this window the machine will then wait until it is inside of this window to reboot.
+### Configure a post-install update hook
 
-For more information about the supported syntax, refer to the [Locksmith documentation][reboot-windows].
+Sometimes you may want to run a custom action after update-engine wrote the new partition out.
+You can create a `/usr/share/oem/bin/oem-postinst` script that gets  two arguments passed.
+The first argument is the slot (`A` or `B`), the second is the temporary mount point where the new `/usr` partition contents can be accessed.
```
krnowak, line 222: I suppose that adding the script is best done with ignition? Also when the post-install hook is run? Right after the installation and before the reboot? Is the script run as root or core or some other user?

pothos: Thanks for the suggestion, I added it

krnowak (approved):
Thanks, looks good.

PR #193, opened 2021-12-16
config-transpiler/dynamic-data.md: correct custom platform guide

pothos:
The custom platform guide talks about using a drop-in unit but when a
non-OEM image is used, it doesn't have the service file at all.
The example used there and in the VMware docs was also lacking some
safety net for temporary malfunctioning metadata servers.

Instruct that the service file should be created instead of using a
drop-in (For existing OEM service files it will work as well to create
the service file under `/etc` which has a higher priority).
Also add some more options to make the oneshot unit more robust.

## How to use

Fixes https://github.com/flatcar-linux/Flatcar/issues/573

### DIFF
--- a/docs/provisioning/config-transpiler/dynamic-data.md
+++ b/docs/provisioning/config-transpiler/dynamic-data.md
@@ -10,6 +10,8 @@ aliases:
 
 Sometimes it can be useful to refer to data in a Container Linux Config that isn't known until a machine boots, like its network address. This can be accomplished with [afterburn][afterburn] (previously called `coreos-metadata`). Afterburn is a very basic utility that fetches information about the current machine and makes it available for consumption. By making it a dependency of services which requires this information, systemd will ensure that coreos-metadata has successfully completed before starting these services. These services can then simply source the fetched information and let systemd perform the environment variable expansions.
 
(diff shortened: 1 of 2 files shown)


docs/provisioning/config-transpiler/dynamic-data.md
```diff
@@ -10,6 +10,8 @@ aliases:
 
 Sometimes it can be useful to refer to data in a Container Linux Config that isn't known until a machine boots, like its network address. This can be accomplished with [afterburn][afterburn] (previously called `coreos-metadata`). Afterburn is a very basic utility that fetches information about the current machine and makes it available for consumption. By making it a dependency of services which requires this information, systemd will ensure that coreos-metadata has successfully completed before starting these services. These services can then simply source the fetched information and let systemd perform the environment variable expansions.
 
+While the `coreos-metadata.service` runs afterburn, it will not set the hostname. The hostname is either through an OEM agent or for particular platforms through afterburn in the initramfs. If afterburn supports your platform and is not invoked in the initramfs by default, you can run it later to set the hostname (`--hostname=/etc/hostname`).
```
krnowak, line 13: Missing verb.

```suggestion
While the `coreos-metadata.service` runs afterburn, it will not set the hostname. The hostname is set either through an OEM agent or for particular platforms through afterburn in the initramfs. If afterburn supports your platform and is not invoked in the initramfs by default, you can run it later to set the hostname (`--hostname=/etc/hostname`).
```

krnowak (changes requested):
Looks good, but had one nit though.

docs/provisioning/config-transpiler/dynamic-data.md
pothos: Thanks!

krnowak (approved):

PR #194, opened 2021-12-21
docs/container/cgroup: add AWS ECS agent as known issue

tormath1:
Signed-off-by: Mathieu Tortuyaux <mathieu@kinvolk.io>

Related-to: https://github.com/flatcar-linux/Flatcar/issues/585

### DIFF
--- a/docs/container-runtimes/switching-to-unified-cgroups.md
+++ b/docs/container-runtimes/switching-to-unified-cgroups.md
@@ -119,5 +119,10 @@ version = 2
  
 For a more detailed discussion of container runtimes, see the [Kubernetes documentation][kube-runtime-docs].
 
+## Known issues
+
+* `aws/amazon-ecs-agent` does not support `cgroupsv2`: [Flatcar issue](https://github.com/flatcar-linux/Flatcar/issues/585), [AWS issue](https://github.com/aws/containers-roadmap/issues/1535).
+
+
 [kube-cgroup-docs]: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/#migrating-to-the-systemd-driver
 [kube-runtime-docs]: https://kubernetes.io/docs/setup/production-environment/container-runtimes/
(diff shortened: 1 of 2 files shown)


docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -119,5 +119,10 @@ version = 2
  
 For a more detailed discussion of container runtimes, see the [Kubernetes documentation][kube-runtime-docs].
 
+## Known issues
+
+* `aws/amazon-ecs-agent` does not support `cgroupsv2`: https://github.com/flatcar-linux/Flatcar/issues/585, https://github.com/aws/containers-roadmap/issues/1535
```
krnowak, line 124: Maybe prettify the links a bit?

```suggestion
* `aws/amazon-ecs-agent` does not support `cgroupsv2`: [Flatcar issue](https://github.com/flatcar-linux/Flatcar/issues/585), [AWS issue](https://github.com/aws/containers-roadmap/issues/1535).
```

krnowak (commented):
Should merging this PR indeed close the issue in flatcar?

tormath1:
> Should merging this PR indeed close the issue in flatcar?

If I understand correctly, Flatcar does **not** ship `aws-ecs-agent` right ? So I guess having this issue referenced as "known issue" is the best we can do at the moment (except contribution to `aws-ecs-agent` to bring cgroupsv2 support of course).

I don't mind to keep it opened while it's being implemented though. :)

krnowak:
> > Should merging this PR indeed close the issue in flatcar?
> 
> If I understand correctly, Flatcar does **not** ship `aws-ecs-agent` right ? So I guess having this issue referenced as "known issue" is the best we can do at the moment (except contribution to `aws-ecs-agent` to bring cgroupsv2 support of course).
> 

Ah, right. I confused it with the `app-emulation/amazon-ssm-agent` that we have (albeit quite old). Nevermind.

> I don't mind to keep it opened while it's being implemented though. :)

Hm. I dunno actually. But some kind of a reminder would be useful when it gets implemented, so we can drop this "known issue".

krnowak (approved):

PR #195, opened 2021-12-22
reference/SDK: rework SDK docs,cover container SDK

t-lo:
This PR updates the developer reference documentation to cover the SDK container.
It also repurposes the "building production images" document to discuss containerised builds and CI automation.

### DIFF
--- a/docs/reference/developer-guides/_index.md
+++ b/docs/reference/developer-guides/_index.md
@@ -5,10 +5,12 @@ aliases:
     - ../os/developer-guides
 ---
 
-Most users will never have to build Flatcar Container Linux from source or modify it in any way. If you have a need to modify Flatcar Container Linux, we provide an SDK that allows you to build your own developer images. We also provide OEM functionality for cloud providers and other companies that must customize Flatcar Container Linux to run within their environment.
+This section is aimed at curious developers interested in building Flatcar Container Linux from source and/or in modifying the OS.
+We provide a containerised SDK that allows you to extend Flatcar and to build your own OS images.
+We also provide OEM functionality for cloud providers and similar use cases to customize Flatcar Container Linux to run within their environment.
 
 * [Guide to building custom Flatcar images from source][mod-cl]
-* [Building production images][production-images]
+* [Vending production images / CI integration][production-images]
 * [Building custom kernel modules][kernel-modules]
 * [SDK tips and tricks][sdk-tips]
 * [SDK build process][sdk-bootstrapping]
@@ -21,4 +23,4 @@ Most users will never have to build Flatcar Container Linux from source or modif
 [mod-cl]: sdk-modifying-flatcar
 [kernel-modules]: kernel-modules
 [sdk-bootstrapping]: sdk-bootstrapping
-[mantle-utils]: https://github.com/kinvolk/mantle/blob/flatcar-master/README.md#kola
+[mantle-utils]: https://github.com/flatcar-linux/mantle/blob/flatcar-master/README.md#kola
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -5,22 +5,18 @@ aliases:
     - ../../os/sdk-building-production-images
 ---
 
-## This document is not maintained
-
-This document still contains useful pointers, but the details are not necessarily up to date.
+FIXME TODO: update to match container builds
 
 ## Introduction
 
-In general the automated process should always be used but in a pinch putting together a release manually may be necessary. All release information is tracked in the [manifest][flatcar-manifest] git repository which is usually organized like so:
+This guide discusses the OS build process and is aimed at audiences comfortable with cutting their very own Flatcar releases. For this purpose we'll have a closer look at the CI automation stubs provided in the [scripts repository][scripts-repo-ci].
 
-* build-109.xml (previous release manifest)
-* build-115.xml (current release manifest)
-* master.xml    (master branch manifest)
-* version.txt   (current version information)
-* default.xml -> master.xml
-* release.xml -> build-115.xml
+It is assumed that readers are familiar with the [SDK][mod-cl] and the general build process outlined in the  [CI automation][scripts-repo-ci].
 
-[flatcar-manifest]: https://github.com/kinvolk/manifest
+
+FIXME TODO: update to match container builds
+
+
 
 ## Tagging releases
 
@@ -70,3 +66,7 @@ The generated production image is bootable as-is by qemu but for a larger ROOT p
 ## Tips and Tricks
 
 We've compiled a [list of tips and tricks](sdk-tips-and-tricks) that can make working with the SDK a bit easier.
+
+
+[scripts-repo-ci]: https://github.com/flatcar-linux/scripts/ci-automation
+[mod-cl]: sdk-modifying-flatcar
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -95,8 +95,8 @@ In that case, the above 3 steps are preceded by
 
 1. Compile all core and SDK packages from source to generate a new SDK root FS and build a tarball from that; generate a new SDK release version and set the OS release to the same (SDK) version.
 2. Build a base SDK container image using the tarball from 1.
-   a. build amd64 and arm64 toolchains and related board support
-   b. then, from the image from 2.a, generate from scratch 3 container images - "all", "amd64", and "arm64" with the respective board support included.
+   1. build amd64 and arm64 toolchains and related board support
+   2. then, from the image from 2. i., generate from scratch 3 container images - "all", "amd64", and "arm64" with the respective board support included.
 
 
 Running all 5 steps in one go will produce a new SDK and new OS image based on that new SDK.
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -68,5 +68,5 @@ The generated production image is bootable as-is by qemu but for a larger ROOT p
 We've compiled a [list of tips and tricks](sdk-tips-and-tricks) that can make working with the SDK a bit easier.
 
 
-[scripts-repo-ci]: https://github.com/flatcar-linux/scripts/ci-automation
+[scripts-repo-ci]: https://github.com/flatcar-linux/scripts/tree/main/ci-automation
 [mod-cl]: sdk-modifying-flatcar
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -5,68 +5,140 @@ aliases:
     - ../../os/sdk-building-production-images
 ---
 
-FIXME TODO: update to match container builds
 
 ## Introduction
 
-This guide discusses the OS build process and is aimed at audiences comfortable with cutting their very own Flatcar releases. For this purpose we'll have a closer look at the CI automation stubs provided in the [scripts repository][scripts-repo-ci].
+This guide discusses automating the OS build process and is aimed at audiences comfortable with producing, testing, and distributing their very own Flatcar releases. For this purpose we'll have a closer look at the CI automation stubs provided in the [scripts repository][scripts-repo-ci].
 
-It is assumed that readers are familiar with the [SDK][mod-cl] and the general build process outlined in the  [CI automation][scripts-repo-ci].
-
-
-FIXME TODO: update to match container builds
+It is assumed that readers are familiar with the [SDK][mod-cl] and the general build process outlined in the [CI automation][scripts-repo-ci].
 
 
+## Stabilisation process and versioning
 
-## Tagging releases
+The Flatcar OS version number follows the pattern MMMM.m.p[ppp] - "M" being the major number, "m" the minor, and "p" the patch level.
+Specifically:
+- A new major version number is introduced with every new Alpha release.
+- The minor version denotes the stabilisation level; 0 means Alpha, 1 is Beta, and 2 is Stable.
+- The patch level denotes incremental releases within the same channel and is incremented e.g. to address issues before promoting a major version to the next stabilisation phase.
 
-The first step of building a release is updating and tagging the release in the manifest git repository. A typical release off of master involves the following steps:
+Roughly, every second Alpha major release goes Beta, and every second Beta major release goes stable.
+A notable exception is the "3033" major release used in the example below; this release shipped ARM64 support and was moved to Stable faster than usual.
+This allows for swift iterations in the "Alpha" channel while keeping Stable major releases ... well ... stable, and ensuring new major Stable releases introduce meaningful sets of updates and new features.
 
- 1. Make sure you are on the master branch: `repo init -b master`
- 2. Sync/checkout source, excluding local changes: `repo sync --detach`
- 3. In the scripts directory: `./tag_release --push`
-
-That was far too easy, if you need to do it the hard way try this:
-
- 1. Make sure you are on the master branch: `repo init -b master`
- 2. Sync/checkout source, excluding local changes: `repo sync --detach`
- 3. Switch to the somewhat hidden manifests checkout: `cd .repo/manifests`
- 4. Update `version.txt` with the desired version number.
-    * COREOS_BUILD is the major version number, and should be the number of days since July 1st, 2013. COREOS_BRANCH should start at 0 and is incremented for every normal release based on a particular COREOS_BUILD version. COREOS_PATCH is reserved for exceptional situations such as emergency manual releases and should normally be 0.
-    * The complete version string is COREOS_BUILD.COREOS_BRANCH.COREOS_PATCH
-    * COREOS_SDK_VERSION should be the complete version string of an existing build. `cork` uses this to pick what SDK tarball to use when creating a fresh chroot and provides a fallback set of binary packages to use when the current release's packages are unavailable. Usually it will be one release behind COREOS_BUILD.
- 5. Generate a release manifest: `repo manifest -r -o build-$BUILD.xml` where `$BUILD` is the current value of COREOS_BUILD in `version.txt`.
- 6. Update `release.xml`: `ln -sf build-$BUILD.xml release.xml`
- 7. Commit! `git add build-$BUILD.xml; git commit -a`
- 8. Tag! `git tag v$BUILD.$BRANCH.$PATCH`
- 9. Push! `git push origin HEAD:master HEAD:dev-channel HEAD:build-$BUILD v$BUILD.$BRANCH.$PATCH`
-
-If a release branch needs to be updated after master has moved on the procedure is similar. Unfortunately since tagging branched releases (not on master) is a bit trickier to get right the `tag_release` script cannot be used. The automated build will kick off after updating the `dev-channel` branch.
-
- 1. Check out the release instead of master: `repo init -b build-$BUILD -m release.xml`
- 2. Sync, cherry-pick, push, and whatever else is required to publish the desired changes in the repo-managed projects. If the desired changes are already published (such as if you are just updating to a later commit from a project's master branch) then this can be skipped.
- 3. `cd .repo/manifests`
- 4. Update `version.txt` as desired. Usually just increment COREOS_PATCH.
- 5. Update `build-$BUILD.xml` as desired. The output of `repo manifest -r` shouldn't be used verbatim this time because it won't generate meaningful values for the `upstream` project attribute when starting from a release manifest instead of `master.xml` but it can be useful for looking up the git commit to update the `revision` attribute to. If the new git commit is on a branch other than master be sure to update the `upstream` attribute with the appropriate ref spec for that branch.
- 6. If this is the first time this branch has been updated on its own update the `default.xml` link so checking out this manifest branch with repo init but without the `-m` argument works: `ln -sf build-$BUILD.xml default.xml`
- 7. Commit! `git commit -a`
- 8. Tag! `git tag v$BUILD.$BRANCH.$PATCH`
- 9. Push! `git push origin HEAD:dev-channel HEAD:build-$BUILD v$BUILD.$BRANCH.$PATCH`
-
-Now you can start building images! This will build an image that can be ran under KVM and uses near production values.
-
-Note: Add `COREOS_OFFICIAL=1` here if you are making a real release. That will change the version to leave off the build id suffix.
-
-```shell
-./build_image prod --group alpha
+A good way to look at releases and stabilisation through channels is to consider major releases as branches from "main", while Alpha, Beta and Stable releases are distinct points in the lifecycle of a release branch:
+```
+  main
+  ...
+   +-- Alpha-2983.0.0
+   |        +-- Beta-2983.1.0
+   |                +--- Stable-2983.2.0
+   |                +--- Stable-2983.2.1
+   |
+   +-- Alpha-3005.0.0
+   +-- Alpha-3005.0.1
+   |
+   +-- Alpha-3033.0.0
+   |        +-- Beta-3033.1.0
+   |        +-- Beta-3033.1.1
+   |                +--- Stable-3033.2.0
+   |
+   +-- Alpha-3046.0.0
+   |
+   +-- Alpha-3066.0.0
+   |        +-- Beta-3066.1.0
+  ...
 ```
 
-The generated production image is bootable as-is by qemu but for a larger ROOT partition or VMware images use `image_to_vm.sh` as described in the final output of `build_image`.
 
-## Tips and Tricks
+## On versioning
 
-We've compiled a [list of tips and tricks](sdk-tips-and-tricks) that can make working with the SDK a bit easier.
+For Flatcar versioning, the scripts repo is authoritative: 
+Versioning is controlled by the [`version.txt` file in the scripts repo](https://github.com/flatcar-linux/scripts/blob/main/sdk_container/.repo/manifests/version.txt) as well as via the commit pointers of the script repo's `portage-stable` and `coreos-overlay` gitmodules.
+`version.txt` contains version strings for both the SDK version as well as the OS image version.
 
+Core idea is that a simple
+```shell
+git checkout 3033.2.0
+git submodules update --init --reqursive
+```
+will set up the scripts repo for development on top of Flatcar release `3033.2.0`.
+
+Keeping `version.txt`,and submodules in sync, updating version strings, and generating version tags is one of the main concerns of the build automation scripts.
+Running a new build via the CI automation will *always* generate a new version.
+This can be non-production version, e.g. nightly build or a PR or branch build - in which case it should be given an appendix following the `MMMM.mm.pp` number.
+The official Flatcar CI uses `-nightly-YYYYMMDD-hhmm` as appendix for nightly builds, e.g. the tag `alpha-3066.0.0-nightly-20221231-0139` would refer to the nightly of the 31st of December, 2021.
+However, custom CI implementations may freely chose to use different appendices.
+
+
+Version information is a mandatory parameter which the CI implementation must feed into the CI automation scripts. Two of the build steps (detailed on below) take version parameters:
+- The SDK bootstrap takes a version string that is used for both the new SDK as well as the downstream OS image version.
+- The OS packages build step takes a version string that is used for the new OS image version.
+
+Both scripts will, based on a given version string:
+1. check out a respective version tag in both `coreos-overlay` and `portage-stable`
+2. update the `version.txt` file accordingly
+3. create a new commit with the above changes
+4. tag the commit with the version string and push the tag.
+
+
+## Build automation and build steps
+
+The Flatcar Container Linux build process consists of
+
+1. compiling packages from source, and generating a new OS image release version / tag
+2. creating a generic OS image file from the resulting binary packages
+3. creating one or more vendor-specific image files from the generic OS image.
+
+
+Optionally, the build process may include building the SDK from scratch based on a previous - exisiting - SDK, e.g. to update core build tools and utilities.
+In that case, the above 3 steps are prepended by
+
+1. Compile all core and SDK packages from source to generate a new SDK root FS and build a tarball from that; generate a new SDK release version and set the OS release to the same (SDK) version.
+2. Build a base SDK container image using the tarball from 1.
+   2.a build amd64 and arm64 toolchains and related board support
+   2.b then, from the image from 2.a, generate from scratch 3 container images - "all", "amd64", and "arm64" with the respective board support included.
+
+
+Running all 5 steps in one go will produce a new SDK and new OS image based on that new SDK.
+In this pipeline, both a new SDK version as well as a new OS image version are generated.
+This is what we call a "full" (or "full-all") build.
+The main use case is for nightly builds of the "main" branches where development of new features happen.
+A new major version release will also use this process (and can be seen as a "special case" of a nightly build of "main").
+New major releases always include a new SDK.
+
+Running only the 3 OS image steps is used for active (i.e. supported) release branches.
+This uses an existing SDK and thus only generates a new OS image version.
+Usually, stabilisation of a major release (alpha -> beta -> stable) uses the same SDK release during its lifetime, so there's no need to always build the SDK.
+Only in rare cases it is necessary to update the SDK after a new major version has been published.
+
+
+### Automation scripts
+
+The [build automation scripts][scripts-repo-ci] reflect the 5 steps outlined above; each step is done in a separate script.
+Check out the build automation's README.md to get an overview.
+Each of the scripts contains documentation of the INPUTs and OUTPUTs of the respective build step:
+
+1. [`sdk_bootstrap.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/sdk_bootstrap.sh) builds a new SDK tarball from scratch
+2. [`sdk_container.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/sdk_container.sh) builds an SDK container image from a tarball
+3. [`packages.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/packages.sh) builds all binary packages for an OS image
+4. [`image.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/image.sh) builds a generic OS image
+5. [`vms.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/vms.sh) builds vendor-specific images
+
+CI / build automation infrastructure should set up the steps in a build pipeline.
+Artifacts of a preceding build step is fed into the succeeding step.
+The scripts are meant to implement build logic in a CI agnostic manner; the concrete CI system used (Jenkins, Bamboo, etc.) should require only very minimal glue logic to run builds.
+
+The build scripts should run on most Linux-based nodes out of the box; `git` and `docker` are the only requirements.
+In the Flatcar project, we use Flatcar Container Linux on our CI worker nodes.
+
+
+### Auxiliary infrastructure
+
+Apart from infrastructure to run the CI /builds on we also need a server for caching build artifacts.
+Build artifacts are mostly container images - with only few exceptions - and are almost always huge (some gigabites).
+To not overly pollute CI workers' disk space, the build scripts support an "artifact cache" server.
+Requirements for this server are rather simple - it should have sufficient disk space (we use 7TB on Flatcar's CI and can hold ~50 past builds), ssh access (for rsync) and serve artifacts from the (rsync/ssh) path prefix via HTTPS.
+See the `BUILDCACHE_...` settings in the [CI automation settings file](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci-config.env) for adapting the build scripts to your environment.
 
 [scripts-repo-ci]: https://github.com/flatcar-linux/scripts/tree/main/ci-automation
 [mod-cl]: sdk-modifying-flatcar
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -141,4 +141,4 @@ Requirements for this server are rather simple - it should have sufficient disk 
 See the `BUILDCACHE_…` settings in the [CI automation settings file](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci-config.env) for adapting the build scripts to your environment.
 
 [scripts-repo-ci]: https://github.com/flatcar-linux/scripts/tree/main/ci-automation
-[mod-cl]: sdk-modifying-flatcar.md
+[mod-cl]: sdk-modifying-flatcar
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -59,7 +59,7 @@ Versioning is controlled by the [`version.txt` file in the scripts repo](https:/
 Core idea is that a simple
 ```shell
 git checkout 3033.2.0
-git submodules update --init --recursive
+git submodule update --init --recursive
 ```
 will set up the scripts repo for development on top of Flatcar release `3033.2.0`.
 
--- a/docs/reference/developer-guides/sdk-building-production-images.md
+++ b/docs/reference/developer-guides/sdk-building-production-images.md
@@ -23,7 +23,7 @@ Specifically:
 
 Roughly, every second Alpha major release goes Beta, and every second Beta major release goes stable.
 A notable exception is the "3033" major release used in the example below; this release shipped ARM64 support and was moved to Stable faster than usual.
-This allows for swift iterations in the "Alpha" channel while keeping Stable major releases ... well ... stable, and ensuring new major Stable releases introduce meaningful sets of updates and new features.
+This allows for swift iterations in the "Alpha" channel while keeping Stable major releases … well … stable, and ensuring new major Stable releases introduce meaningful sets of updates and new features.
 
 A good way to look at releases and stabilisation through channels is to consider major releases as branches from "main", while Alpha, Beta and Stable releases are distinct points in the lifecycle of a release branch:
 ```
@@ -59,15 +59,15 @@ Versioning is controlled by the [`version.txt` file in the scripts repo](https:/
 Core idea is that a simple
 ```shell
 git checkout 3033.2.0
-git submodules update --init --reqursive
+git submodules update --init --recursive
 ```
 will set up the scripts repo for development on top of Flatcar release `3033.2.0`.
 
-Keeping `version.txt`,and submodules in sync, updating version strings, and generating version tags is one of the main concerns of the build automation scripts.
+Keeping `version.txt` and submodules in sync, updating version strings, and generating version tags is one of the main concerns of the build automation scripts.
 Running a new build via the CI automation will *always* generate a new version.
-This can be non-production version, e.g. nightly build or a PR or branch build - in which case it should be given an appendix following the `MMMM.mm.pp` number.
-The official Flatcar CI uses `-nightly-YYYYMMDD-hhmm` as appendix for nightly builds, e.g. the tag `alpha-3066.0.0-nightly-20221231-0139` would refer to the nightly of the 31st of December, 2021.
-However, custom CI implementations may freely chose to use different appendices.
+This can be non-production version, e.g. nightly build or a PR or branch build - in which case it should be given a suffix following the `MMMM.mm.pp` number.
+The official Flatcar CI uses `-nightly-YYYYMMDD-hhmm` as suffix for nightly builds, e.g. the tag `alpha-3066.0.0-nightly-20221231-0139` would refer to the nightly of the 31st of December, 2021.
+However, custom CI implementations may freely choose to use different suffixes.
 
 
 Version information is a mandatory parameter which the CI implementation must feed into the CI automation scripts. Two of the build steps (detailed on below) take version parameters:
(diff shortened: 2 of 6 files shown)


docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -70,3 +66,7 @@ The generated production image is bootable as-is by qemu but for a larger ROOT p
 ## Tips and Tricks
 
 We've compiled a [list of tips and tricks](sdk-tips-and-tricks) that can make working with the SDK a bit easier.
+
+
+[scripts-repo-ci]: https://github.com/flatcar-linux/scripts/ci-automation
```
krnowak, line 71: That's 404.

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,181 @@ aliases:
...
-gpg: Good signature from "Flatcar Buildbot (Official Builds) <buildbot@flatcar-linux.org>" [unknown]
-Primary key fingerprint: F88C FEDE FF29 A5B4 D952  3864 E25D 9AED 0593 B34A
-     Subkey fingerprint: 858A 560F 97C9 AEB2 2EC1  C732 961D DDD5 250D 4A42
-```
+Cloning the repo will have it land on the `main` branch, which can be thought of as "alpha-next" - i.e. the next major Alpha release.
+Even though main is smoke-tested in nightly builds, it might occasionally be broken in subtle ways.
+This can make it harder to track down issues introduced by actual changes to Flatcar.
 
-Then proceed with the installation of the cork binary to a location on your path:
+* Release **tags** signify speficic (past) releases, like "stable-2905.2.4" or "beta-3033.1.1". Tags are created in release branches.
```
krnowak, line 80:
```suggestion
* Release **tags** signify specific (past) releases, like "stable-2905.2.4" or "beta-3033.1.1". Tags are created in release branches.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,181 @@ aliases:
...
-To use the SDK chroot in the future, run `cork enter` from the directory you installed the SDK to (i.e. where you ran `cork create`).
 
-## Building an image
+#### A note on persistence
 
-### A brief introduction to Gentoo
+`run_sdk_container` re-uses containers once started; containers to be re-used are identified by name (see above). 
+Persistence helps with keeping changes in your work environment across container runs.
+**Keep in mind though that a new container will be created if the working commit in the scripts repository changes**.
+This is usually derired to prevent version muddling.
```
krnowak, line 153:
```suggestion
This is usually desired to prevent version muddling.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +253,132 @@ You can log in via SSH keys or with a different ssh port by running this example
...
+Each package directory may contain ebuild files for multiple versions, e.g. `dev-lang/python` contains a host of python versions (used in the SDK).
+Furthermore, each package directory contains a `Manifest` file with cryptographic checksums and file sizes of the package's source tarball(s), and may contain a `files/` directory containing auxiliary files to build / install the package, e.g. patches or config files.
+
+Multiple package sources - in separate directories - can be stacked on top of each other.
+These “overlays” allow custom extensions or even custom sub-trees on top of an existing foundation.
+In these stacks, “upper” level packages override “lower” level ones.
+The Flatcar build system uses a fork of Gentoo upstream’s [portage-stable](https://github.com/flatcar-linux/portage-stable) as its base, and the overlay repository [coreos-overlay](https://github.com/flatcar-linux/coreos-overlay) for Flatcar specific modifications and packages on top.
+
+Packages are built using "ebuild" files.
+These files contain a package’s dependencies - both build and runtime - as well as implement callbacks for downloading, patching, building, and installing a package.
```
krnowak, line 279:
```suggestion
These files contain package’s dependencies - both build and runtime - as well as implement callbacks for downloading, patching, building, and installing a package.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -348,19 +387,9 @@ Lastly, the SDK might lack unmasks if the respective architecture is masked in t
...
-which will print the changes and prompt you before writing them. Run `emerge-amd64-usr --newuse <group>/<package>` again to proceed.
-
-**NOTE**: a bug in the Chromium part of the SDK will lead to the following warning:
-```shell
---- Invalid atom in /build/amd64-usr/etc/portage/package.unmask/cros-workon:
-     =<group>/<package>
-```
-The SDK creates softlinks for both `/build/amd64-usr/etc/portage/package.unmask/cros-workon` and `/build/amd64-usr/etc/portage/package.keywords/cros-workon` to the same file - to `.config/cros_workon/amd64-usr` (outside the SDK chroot). This effectively merges all required keywords and unmasks into the same file. The keywords and unmasks files follow different semantics, leading to emerge printing the warning above. Since invalid atoms will be ignored, it’s safe to ignore the warning for the time being.
+To proceed, remove the architecture mask in the ebuild file (remove the `~` - for example, change `~arm64` to `arm64`) and re-run emerge.
+FLatcar follows its own stabilisation process (through the Alpha - Beta - Stable channels); it's perfectly fine to unmask a package upstream considers unstable.
```
krnowak, line 392:
```suggestion
Flatcar follows its own stabilisation process (through the Alpha - Beta - Stable channels); it's perfectly fine to unmask a package upstream considers unstable.
```

krnowak (commented):
Found some typos.

docs/reference/developer-guides/sdk-building-production-images.md
t-lo, line 71:
```suggestion
https://github.com/flatcar-linux/scripts/tree/main/ci-automation
```

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
- 9. Push! `git push origin HEAD:master HEAD:dev-channel HEAD:build-$BUILD v$BUILD.$BRANCH.$PATCH`
+Core idea is that a simple
+```shell
+git checkout 3033.2.0
+git submodules update --init --reqursive
+```
+will set up the scripts repo for development on top of Flatcar release `3033.2.0`.
 
-If a release branch needs to be updated after master has moved on the procedure is similar. Unfortunately since tagging branched releases (not on master) is a bit trickier to get right the `tag_release` script cannot be used. The automated build will kick off after updating the `dev-channel` branch.
+Keeping `version.txt`,and submodules in sync, updating version strings, and generating version tags is one of the main concerns of the build automation scripts.
```
pothos, line 66: What commands should be used to set the new submodule references? I want to add them to https://github.com/flatcar-linux/scripts/blob/main/tag_release to make sure we update them from now on even if we didn't make the switch.

t-lo: There are library functions for this; should be straightforward. We need 4 or 5 inputs depending on what we are releasing. Please note that the functions only care about the `scripts` repo so they expect either commit IDs or tags that already exist in these repos. Furthermore, the scripts only handle tagging, not branching. It is expected that the `scripts` repo is already on the correct branch.

For new major releases:
1. coreos-overlay commit-ish, tag, or branch
2. portage-stable commit-ish, tag, or branch
3. new SDK version number (to be used for both SDK and OS)
4. new release version tag name (channel + version number)

For channel propagation and other incremental releases that do not introduce a new major version we additionally need
4. new OS image version number

In the scripts repo, run
```shell
source sdk_lib/sdk_container_common.sh
source ci-automation/ci_automation_common.sh

update_submodules <coreos-overlay commit-ish> <portage-stable commit-ish>

create_versionfile <SDK-version> [<os-version>]

update_and_push_version <release-tag>
```

e.g. for a new Alpha-3110 based on the current main branches:
```shell
source sdk_lib/sdk_container_common.sh
source ci-automation/ci_automation_common.sh

update_submodules main main
create_versionfile 3110.0.0
update_and_push_version alpha-3110.0.0
```

Or for a Beta-3110.1.0 based on the above Alpha:
```shell
source sdk_lib/sdk_container_common.sh
source ci-automation/ci_automation_common.sh

update_submodules flatcar-3110 flatcar-3110
create_versionfile 3110.1.0
update_and_push_version beta-3110.1.0
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
krnowak: I'd say that the "a" article is bound to the "dependencies" word (and since "dependencies" is plural, it shouldn't have any indefinite article), not to "package's". But not an expert on articles. Anyway, I wouldn't go with "packages' dependencies", because the sentence is about a single package. So how about:

```suggestion
These files contain dependencies of a package - both build and runtime - as well as implement callbacks for downloading, patching, building, and installing the package.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -270,6 +222,9 @@ $ ./image_to_vm.sh --from=../build/images/arm64-usr/developer-latest [--board=..
 For other vendor images, pass the `--format=` parameter (see `./image_to_vm.sh --help`).
 In general, `image_to_vm.sh` will read the generic disk image, install any vendor specific tools to the OEM partition where applicable (e.g. Azure VM tools for the Azure VM), and produce a vendor specific image. In the case of QEMU, a qcow2 image is produced. QEMU does not require vendor specific tooling in the OEM partition.
 
+On the host outside the container, the image(s) built are located in `__build__/images/...`.
```
krnowak, line 225:
```suggestion
On the host outside the container, the image(s) built are located in `__build__/images/…`.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +253,132 @@ You can log in via SSH keys or with a different ssh port by running this example
 
 Now for the interesting part! We are going to discuss 2 ways of making changes: adding or upgrading a package, and modifying the kernel configuration.
 
-### Add (or update) a package
+### A brief introduction to Gentoo how it relates to the SDK
```
krnowak, line 256: I think an "and" is missing here:

```suggestion
### A brief introduction to Gentoo and how it relates to the SDK
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +253,132 @@ You can log in via SSH keys or with a different ssh port by running this example
...
+
+Flatcar Container Linux is based on ChromiumOS, which is based on Gentoo.
+While the ChromiumOS heritage has faded and is barely visible nowadays, we heavily leverage Gentoo processes and tools.
+
+Contrary to traditional Linux distributions, Gentoo applications and “packages” are compiled at installation time.
+Gentoo itself does not ship packages - instead, it consists of a massive number of ebuild files to build applications at installation time (that’s an oversimplification as there are binary package caches, but that’s beyond the scope of this document).
+While the Flatcar SDK can be understood as a Gentoo derivate, the OS image is special.
```
krnowak, line 263:
```suggestion
While the Flatcar SDK can be understood as a Gentoo derivative, the OS image is special.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +253,132 @@ You can log in via SSH keys or with a different ssh port by running this example
...
-$ cp -R <gentoo-repo-dir>/<group>/<package> <flatcar-SDK>/src/third_party/portage-stable/<group>/
+~/trunk/src/scripts $ git clone --depth 5 https://github.com/gentoo/gentoo.git
+~/trunk/src/scripts $ mkdir -p ../third_party/portage-stable/<group>/
+~/trunk/src/scripts $ cp -R gentoo/<group>/<package> ../third_party/portage-stable/<group>/
+~/trunk/src/scripts $ emerge-amd64-usr --newuse <group>/<package>
+# optional - add missing eclass
+~/trunk/src/scripts $ cp gentoo/eclass/<eclass-name>.eclass ../third_party/portage-stable/eclass/
+~/trunk/src/scripts $ emerge-amd64-usr --newuse <group>/<package>
+# optional - unmask package
+~/trunk/src/scripts $ vim ../third_party/portage-stable/<group>/<package>/<package-version>.ebuild
```
krnowak, line 323: I'd say that this is a "no no". Instead, the preference is to modify the `package.accept_keywords` in coreos-overlay, so the modification won't be lost when we sync the portage-stable with gentoo next time. So maybe:
```suggestion
# optional - unmask package
~/trunk/src/scripts $ vim ../third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -348,19 +387,9 @@ Lastly, the SDK might lack unmasks if the respective architecture is masked in t
...
 
-which will print the changes and prompt you before writing them. Run `emerge-amd64-usr --newuse <group>/<package>` again to proceed.
-
-**NOTE**: a bug in the Chromium part of the SDK will lead to the following warning:
-```shell
---- Invalid atom in /build/amd64-usr/etc/portage/package.unmask/cros-workon:
-     =<group>/<package>
-```
-The SDK creates softlinks for both `/build/amd64-usr/etc/portage/package.unmask/cros-workon` and `/build/amd64-usr/etc/portage/package.keywords/cros-workon` to the same file - to `.config/cros_workon/amd64-usr` (outside the SDK chroot). This effectively merges all required keywords and unmasks into the same file. The keywords and unmasks files follow different semantics, leading to emerge printing the warning above. Since invalid atoms will be ignored, it’s safe to ignore the warning for the time being.
+To proceed, remove the architecture mask in the ebuild file (remove the `~` - for example, change `~arm64` to `arm64`) and re-run emerge.
```
krnowak, line 391: In the same vein as above:
```suggestion
To proceed, add the package name and version, and its masked architectures to the `package.accept_keywords` file inside the `coreos` profile. Which `package.accept_keywords` file should be updated depends on couple factors - whether it is needed for both SDK and OS image or only for SDK or only for OS image, whether it is needed for both AMD64 and ARM64 images, or only for AMD64 or only for ARM64. Please refer to `README.md` in coreos-overlay for a summary about profiles
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -371,142 +400,212 @@ emerge: there are no ebuilds to satisfy "<group>/<package>:=" for /build/amd64-u
...
+
+</td></tr></table>
+
+Next, we’ll look into changing the kernel configuration - e.g. for adding a kernel module or a core kernel feature not shipped with stock Flatcar.
+This will give you a deep dive into the low-level bits of Gentoo's build and packaging system.
+To modify the configuration of a package we will run its individual build steps manually - by use of `ebuild` instead of `emerge`.
+This will allow for pausing after downloading the sources, to change the source tree configuration before building and installing.
+
+Our first step is to set you all up with a pre-configured stock Flatcar Linux kernel to base your modifications on.
+The Flatcar Linux kernel build is split over multiple gentoo ebuild files which all reside in <code>[coreos-overlay/sys-kernel/](https://github.com/kinvolk/coreos-overlay/tree/main/sys-kernel)</code>:
```
krnowak, line 502: Won't this work?

```suggestion
The Flatcar Linux kernel build is split over multiple gentoo ebuild files which all reside in [`coreos-overlay/sys-kernel/`](https://github.com/kinvolk/coreos-overlay/tree/main/sys-kernel):
```

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
- 7. Commit! `git commit -a`
- 8. Tag! `git tag v$BUILD.$BRANCH.$PATCH`
- 9. Push! `git push origin HEAD:dev-channel HEAD:build-$BUILD v$BUILD.$BRANCH.$PATCH`
 
-Now you can start building images! This will build an image that can be ran under KVM and uses near production values.
+## On versioning
 
-Note: Add `COREOS_OFFICIAL=1` here if you are making a real release. That will change the version to leave off the build id suffix.
+For Flatcar versioning, the scripts repo is authoritative: 
+Versioning is controlled by the [`version.txt` file in the scripts repo](https://github.com/flatcar-linux/scripts/blob/main/sdk_container/.repo/manifests/version.txt) as well as via the commit pointers of the script repo's `portage-stable` and `coreos-overlay` gitmodules.
```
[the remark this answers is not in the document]
t-lo: Absolutely. I just left it there to get a first iteration of the containerised SDK / versioned git submodules out. One of the design goals of that first iteration was to reproduce *exactly* what users are used to from the `cork` SDK, which included putting the versionfile to a well-known place.
We can absolutely move it elsewhere in a follow-up operation; I feel there's a lot of room for more simplifications after we have established the groundwork.

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
+This uses an existing SDK and thus only generates a new OS image version.
+Usually, stabilisation of a major release (alpha -> beta -> stable) uses the same SDK release during its lifetime, so there's no need to always build the SDK.
+Only in rare cases it is necessary to update the SDK after a new major version has been published.
+
+
+### Automation scripts
+
+The [build automation scripts][scripts-repo-ci] reflect the 5 steps outlined above; each step is done in a separate script.
+Check out the build automation's README.md to get an overview.
+Each of the scripts contains documentation of the INPUTs and OUTPUTs of the respective build step:
```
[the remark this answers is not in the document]
t-lo: It's all-caps in the scripts' documentation, but I agree to user lowercase here for readability.

docs/reference/developer-guides/sdk-tips-and-tricks.md
```diff
@@ -194,6 +151,24 @@ If you want to permanently leave you can run the following:
 crossystem disable_dev_request=1; reboot
 ```
 
+## Re-initialise the SDK container
+
+By default, the SDK container is re-used when using the `./run_sdk_container` script; all your changes within the container are preserved.
+To reset the container, list all docker containers:
+```shell
+docker ps --all
+...
```
t-lo, line 160:
```suggestion
…
```

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
-## Tips and Tricks
+Apart from infrastructure to run the CI / builds on we also need a server for caching build artifacts.
+Build artifacts are mostly container images - with only few exceptions - and are almost always huge (some gigabytes).
+To not overly pollute CI workers' disk space, the build scripts support an "artifact cache" server.
+Requirements for this server are rather simple - it should have sufficient disk space (we use 7TB on Flatcar's CI and can hold ~50 past builds), ssh access (for rsync) and serve artifacts from the (rsync/ssh) path prefix via HTTPS.
+See the `BUILDCACHE_…` settings in the [CI automation settings file](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci-config.env) for adapting the build scripts to your environment.
 
-We've compiled a [list of tips and tricks](sdk-tips-and-tricks) that can make working with the SDK a bit easier.
+[scripts-repo-ci]: https://github.com/flatcar-linux/scripts/tree/main/ci-automation
+[mod-cl]: sdk-modifying-flatcar.md
```
t-lo, line 144:
```suggestion
[mod-cl]: sdk-modifying-flatcar
```

docs/reference/developer-guides/sdk-tips-and-tricks.md
```diff
@@ -312,3 +288,7 @@ e96281a6-d1af-4bde-9a0a-97b76e56dc57
 - Flatcar Container Linux Root: 5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6
 - Flatcar Container Linux Reserved: c95dc21a-df0e-4340-8d7b-26cbfa9a03e0
 - Flatcar Container Linux Raid Containing Root: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+
+
+
+[mod-cl]: sdk-modifying-flatcar.md
```
t-lo, line 294:
```suggestion
[mod-cl]: sdk-modifying-flatcar
```

t-lo (commented):
Need to revert `.md` fixes to documentation URIs since since the final, rendered docs don't use `.md` in the URI.

krnowak:
> Need to revert `.md` fixes to documentation URIs since since the final, rendered docs don't use `.md` in the URI.

Meh. Maybe the renderer could be made smarter by dropping those `.md` extensions. Anyway, not a part of this PR, but thanks for letting me know.

docs/reference/developer-guides/sdk-modifying-flatcar.md
t-lo: Agreed, makes sense - in particular since we're striving for full alignment and potential deprecation of portage-stable.

pothos:
More a meta comment: when is a good point to merge this? Currently we didn't switch over to it in Jenkins nor do the releases have the right submodule commits pinned, I think at least the submodule references updates should be automated for the main and maintenance branches and done for the release tags.

pothos:
> Updating the submodule pinning as well as tagging releases is already automated - and used every night for the containerised nightlies - we just need to start using it for releases.

The `flatcar-3033` branch on scripts has an outdated reference for the coreos-overlay submodule, it looks like we need add the missing automation for this first.

t-lo:
> > Updating the submodule pinning as well as tagging releases is already automated - and used every night for the containerised nightlies - we just need to start using it for releases.
> 
> The `flatcar-3033` branch on scripts has an outdated reference for the coreos-overlay submodule, it looks like we need add the missing automation for this first.

Yes, because it's a branch, and the containerised CI automation only cares about tags. So if you e.g. check out the tag `stable-3033.2.0` then you'll get coreos-overlay and portage-stable in the versions released with 3033.2.0.

The nightly builds in the CI also pull the latest submodule branches before tagging, but nightly tags aren't part of a branch. So if you e.g. check out `stable-3033.2.0-nightly-20220104-0139` you will get the state of tonight's `flatcar-3033` branches. These tags (and the respective commits) aren't part of any branch (only the tags are pushed) to prevent nightlies from polluting the scripts branches with new commits every night. 

If we're discussing auto-updating the *branches*, this doesn't need to be a CI automation thing but could even be a github action in the scripts repo: any update of a branch in either coreos-overlay or portage-stable would cause the respective branch of the same name in scripts (if it exists) to automatically update its submodule, copying the submodule's commit message.

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,181 @@ aliases:
...
-Still, during the compilation phase many package builds perform `configure` tests by running a compiled binary or invoke compiled helper binaries.
-This requires the host system to use binary translation (software-based virtualisation) for seamlessly running arm64 binaries.
+At the time of writing the SDK supports two target architectures: AMD64 (x86-64) and ARM64.
+The target architecture can be specified by use of the `--board=` parameter to both `build_packages` and `build_image`:
+* `--board=amd64-usr` will build an x86 image
+* `--board=arm64-usr` will build and ARM64 image
 
-On a Fedora/Debian host system there is a `qemu-user-static` package that sets everything up.
-If you have troubles or use a different host system, check that a statically compiled aarch64 qemu-user binary is referenced with the `:F` flag for pinning it into the kernel so that it is not needed in every chroot:
+If no architecture is specified then AMD64 will be used by default.
```
pothos, line 181: Please, let's keep these two sentences about the distro package, the new NOTE below is too short and sounds like the user would have to just create this file.

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -228,28 +189,19 @@ $ sudo systemctl restart systemd-binfmt.service
...
+It's likely this won't *actually* build the packages but rather download pre-built packages from the Flatcar binary package cache.
+The package cache is updated on every release.
 
 ```shell
 $ ./build_packages [--board=...]
 ```
 
-You only need to use `--board` if you did not use the `--default` option with `setup_board` above, or if you want to build for an architecture that's not your default one.
-The command should download most packages from our binary cache - speeding up the "build" - since we are basing this on an existing release. All packages will be installed to `/build/<arch>`.
+The command should download most packages from our binary cache - speeding up the "build" - since we are basing this on an existing release.
```
pothos, line 201: I think this needs more info on how to make sure that it is set up. In the tips and tricks all the `set_version` commands are removed and the user is left without any info on how to set the package cache up, be it for the maintenance branch or the main branch.

t-lo: They should not need to since information on the binary caches is included in the SDK container, which itself is versioned. `set_version` disappeared because it is replaced (implicitly) by checking out the correct branch or tag of the scripts repo. 

It works like this:
1. check out version A, update submodules
2. `./run_sdk_container` will read the versionfile, use the correct SDK container image for version A
3. `./build_packages` in the container will use the correct bincache for version A.

Now, to switch versions, one would:
1. check out version B, update submodules
2. `./run_sdk_container` now uses a different SDK container image - for version B
3. `./build_packages` in the container will use the bincache for version B.

Additionally, the containers for version A and B will be named accordingly (the container name contains OS and SDK version) - so one could switch back and forth between versions by just checking out the respective `scripts` branch, and never lose intermediate build artifacts (as long as the container is not deleted by the user).

t-lo:
Addressed @pothos feedback:
-  add documentation on ensuring the [release branch tips of submodules are current](https://github.com/flatcar-linux/flatcar-docs/pull/195#issuecomment-1004886030)
- make qemu / binfmt-misc documentation more elaborate and prominent for cross-compiling.

docs/reference/developer-guides/sdk-modifying-flatcar.md
t-lo: Is there something I'm missing that `set_version` does which couldn't be done by using the plain versionfile from `scripts`?

pothos: Ok, so the idea would be to publish nightly SDK containers for the main and the maintenance branches? Maybe without building the SDK since it doesn't need to be rebuilt for maintenance branches?

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
-Note: Add `COREOS_OFFICIAL=1` here if you are making a real release. That will change the version to leave off the build id suffix.
+For Flatcar versioning, the scripts repo is authoritative: 
+Versioning is controlled by the [`version.txt` file in the scripts repo](https://github.com/flatcar-linux/scripts/blob/main/sdk_container/.repo/manifests/version.txt) as well as via the commit pointers of the script repo's `portage-stable` and `coreos-overlay` gitmodules.
+`version.txt` contains version strings for both the SDK version as well as the OS image version.
 
+Core idea is that a simple
 ```shell
-./build_image prod --group alpha
+git checkout 3033.2.0
+git submodules update --init --recursive
```
pothos, line 62:
```suggestion
git submodule update --init --recursive
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,210 @@ aliases:
...
-This requires the host system to use binary translation (software-based virtualisation) for seamlessly running arm64 binaries.
+### Select the target architecture
 
-On a Fedora/Debian host system there is a `qemu-user-static` package that sets everything up.
-If you have troubles or use a different host system, check that a statically compiled aarch64 qemu-user binary is referenced with the `:F` flag for pinning it into the kernel so that it is not needed in every chroot:
+**NOTE on cross-compilation**: if you are cross-compiling make sure a static aarch64 qemu is set up via binfmt-misc on your host machine.
+Some packages compile and execute intermediate commands during their build process - this can break cross-compiling since the commands are built for the target architecture.
+The qemu binary on the host needs to be a static binary since it will be called from within the container context.
+Check if your distro has support for aarch64 in `binfmt-misc` already; on e.g. Fedora there's an `qemu-aarch64` entry in `/proc` for that (the name of the proc file may vary across distributions though):
```
pothos, line 200:
```suggestion
Check if your distro has a `qemu-user-static` package that you can install or whether it has support for aarch64 in `binfmt-misc` already; on e.g. Fedora there's an `qemu-aarch64` entry in `/proc` for that (the name of the proc file may vary across distributions though):
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
t-lo: Yes, that would be the idea. The containerised nightly job for main already builds the SDK container and credentials for pushing from CI to ghcr.io are in place (but currently unused) - it's a one-line change to publish after building.

Maintenance (release) branches don't need a new SDK (made a mistake in my first comment and updated it); the versionfile should determine the bincache URL.

pothos: While this is not done I suggest to keep the `set_version` docs around for getting binary packages when working on the main/maintenance branches

pothos:
This effectively removes the cork docs which to me means that we should try to address the remaining pain points before the users run into them (the last hindrance due to the GCS bucket issue blocked people for a month or so and now we can be happy that all is calm again).

Beside the mentioned nightly binary package topic and the automation for release tagging, I think we should also consider Podman support, even if it's just by having a big warning in the docs that this here won't work when the Fedora `docker` wrapper script for Podman is used.

t-lo:
> This effectively removes the cork docs which to me means that we should try to address the remaining pain points before the users run into them (the last hindrance due to the GCS bucket issue blocked people for a month or so and now we can be happy that all is calm again).

Do we expect users to base their work on `main`? The main branch is regularly broken and that will prevent folks from even getting a base image built locally (i.e. w/o modifications). For release / maintenance branches, packages are available on [mirror.flatcar-linux](https://mirror.release.flatcar-linux.net/).

If we want to make working on the main branch (and using nightlies) more accessible we might want to provide an `alpha-next` binary package cache via [mirror.flatcar-linux](https://mirror.release.flatcar-linux.net/) and wire the nightly builds to use that.

> Beside the mentioned nightly binary package topic and the automation for release tagging, I think we should also consider Podman support, even if it's just by having a big warning in the docs that this here won't work when the Fedora `docker` wrapper script for Podman is used.

IIRC there wasn't much change needed to make `podman` work so why don't we just detect `podman` and add a respective option in [`run_sdk_container`](https://github.com/flatcar-linux/scripts/blob/main/run_sdk_container#L94-L127)?

Tagging commands (with examples) are documented [here](https://github.com/flatcar-linux/flatcar-docs/pull/195#discussion_r777918400).

pothos:
A maintenance branch is something like `flatcar-3033` which I would say also benefits from configuring nightly binary packages from bucket.release.flatcar-linux.net.

docs/reference/developer-guides/sdk-building-production-images.md
pothos: This is not really what we do for the release process, a branch gets created for a new major release first, we don't use `main` to tag the Alpha.
I would say we should not just push a lone tag but actually have it be a commit first on the release maintenance branch.

pothos: I've filed a PR now for our release tooling: https://github.com/flatcar-linux/flatcar-build-scripts/pull/91

pothos:
> IIRC there wasn't much change needed to make `podman` work so why don't we just detect `podman` and add a respective option in [`run_sdk_container`](https://github.com/flatcar-linux/scripts/blob/main/run_sdk_container#L94-L127)?

Draft in https://github.com/flatcar-linux/scripts/pull/199

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,210 @@ aliases:
...
+$ cd scripts
+$ branch="$(git branch -r -l | sed -n 's:origin/\(flatcar-[0-9]\+\)$:\1:p' | sort | tail -n1)"
+$ git checkout "$branch"
+$ git submodule init
+$ git submodule update
+$ for s in sdk_container/src/third_party/coreos-overlay/ sdk_container/src/third_party/portage-stable; do
+$   git -C "$s" checkout "$branch"
+$   git -C "$s" pull --rebase
+$ done
+$ ./run_sdk_container -t
```
pothos, line 37: This fails if it was run once and the `version.txt` got created:
```
Error: error creating container storage: the container name "flatcar-sdk-all-3087.0.0_os-alpha-3066.0.0-52-g1933bf20" is already in use
```

pothos: adding `-n NAME` as mentioned later in the docs gives the same error

pothos: ah, it's the `^/` not working with podman...

t-lo: Interesting - could you please `set -x` at the top of `run_sdk_container` and provide the output? It works here...
(Also - is that with Docker or with Podman?)

pothos: It was with Podman, pushed a fix in https://github.com/flatcar-linux/scripts/pull/199 by leaving the `^/` prefix out

t-lo: Good catch - most interestingly I just added the regex today to fix something else: https://github.com/flatcar-linux/scripts/pull/196

docs/reference/developer-guides/sdk-modifying-flatcar.md
pothos: > the other [solution] is that we can use the new OS version to overwrite the contents in `make.conf`

Thinking again this (overwriting the portage make.conf binhost when the OS version file gets created) is the only real solution because the current behavior is not save for releases either - they could end up reusing the initial major version binary packages instead of rebuilding all of them (didn't check if this is already the case on the nightly test job in jenkins).

t-lo: > Thinking again this (overwriting the portage make.conf binhost when the OS version file gets created) is the only real solution because the current behavior is not save for releases either - they could end up reusing the initial major version binary packages instead of rebuilding all of them (didn't check if this is already the case on the nightly test job in jenkins).

That's a great catch - I'll have a look at adding this to [sdk_entry.sh](https://github.com/flatcar-linux/scripts/blob/main/sdk_lib/sdk_entry.sh) so the container's `make.conf` is always aligned with the versionfile.

t-lo: @pothos https://github.com/flatcar-linux/scripts/pull/201

t-lo:
I'll update this pr / the docs to discuss `checkout` if we can get the PR merged: https://github.com/flatcar-linux/scripts/pull/202

t-lo:
Rebased on latest "main".

t-lo:
Updated to discuss https://github.com/flatcar-linux/scripts/pull/202 (which is now merged) as well as call out podman and docker as supported runtimes.

docs/reference/developer-guides/sdk-building-production-images.md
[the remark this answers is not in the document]
pothos: Maybe a symlink from `/version.txt` could help? For now people will notice the file when it gets modified automatically and shows up red in `git status`

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,195 @@ aliases:
...
-
-
-**You also need a proper git setup**
+<table><tr><td>
 
+**tl;dr** Check out a release branch and start the SDK (this uses the current Alpha release branch).
 ```shell
-$ git config --global user.email "you@example.com"
-$ git config --global user.name "Your Name"
+$ git clone https://github.com/flatcar-linux/scripts.git
```
pothos, line 27:
```suggestion
$ git clone --recurse-submodules https://github.com/flatcar-linux/scripts.git
```
I would always use this in the docs for people to be aware that submodules are needed (even if we have automation to fetch them when they are missing)

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,195 @@ aliases:
...
-
-If you want to build cork from sources instead of using an official release, you'll need golang installed as mantle is written in go.
+The [scripts repository][scripts] - among other things - contains SDK wrapper scripts, a `version.txt` with release version information, and both the [coreos-overlay][coreos] and [portage-stable][portage] ebuild repositories as git submodules (more on ebuilds later).
+A good way to think of the scripts repo is this being Flatcar's "SDK repo".
 
 ```shell
-$ git clone git@github.com:kinvolk/mantle.git
-$ cd mantle
-$ ./build cork
+$ git clone https://github.com/flatcar-linux/scripts.git
```
pothos, line 67:
```suggestion
$ git clone --recurse-submodules https://github.com/flatcar-linux/scripts.git
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -6,220 +6,195 @@ aliases:
...
 ```
-to base your work on the Stable 2905.2.6 release published on October 25th, 2021.
-
-### Enter the SDK
 
-Run
-```shell
-$ cork enter
-```
+This wikk check out the branch or tag in the top-level `scripts` directory and update the submodules, depending on wheter a branch or tag was specified:
```
pothos, line 107:
```suggestion
This will check out the branch or tag in the top-level `scripts` directory and update the submodules, depending on wheter a branch or tag was specified:
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -228,28 +203,26 @@ $ sudo systemctl restart systemd-binfmt.service
...
+* `--board=arm64-usr` will build and ARM64 image
 
-### Build all packages that make up the OS image
+If no architecture is specified then AMD64 will be used by default.
 
-Before we discuss any modifications to the image, we'll do a full image build first. This will create a "known-good" base to mount your changes on.
+### Build the OS image packages
 
-First, build all of the target binary packages:
+It's likely this won't *actually* build the packages but rather download pre-built packages from the Flatcar binary package cache.
```
pothos, line 215:
```suggestion
Beware, it's likely this won't *actually* build the packages but rather download pre-built packages from the Flatcar binary package cache (see below on how to force-rebuild a package that you modified).
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +274,132 @@ You can log in via SSH keys or with a different ssh port by running this example
...
 
-To get access to a rich and up-to-date selection of packages, we’ll use the upstream Gentoo ebuilds repository. We’ll copy the ebuild file of the package we want to add from upstream gentoo to portage-stable, as well as the package’s dependencies.
+### Get to know the SDK chroot
 
-Let’s start by checking out the Gentoo upstream ebuilds to some place outside the SDK. We’ll only do a shallow clone to limit the amount of data we need to download:
+When entering the SDK you are in the `~/trunk/src/scripts` repository which can be seen as the build system.
+It is one of the three repositories that define a Flatcar build:
+1. flatcar-scripts (the directory you're in) contains high-level build scripts to build all packages for an image, to build an image, and to bootstrap an SDK.
```
pothos, line 313:
```suggestion
1. `scripts` (the directory you're in) contains high-level build scripts to build all packages for an image, to build an image, and to bootstrap an SDK.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +274,132 @@ You can log in via SSH keys or with a different ssh port by running this example
...
+~/trunk/src/scripts $ cp gentoo/eclass/<eclass-name>.eclass ../third_party/portage-stable/eclass/
+~/trunk/src/scripts $ emerge-amd64-usr --newuse <group>/<package>
+# optional - unmask package
+~/trunk/src/scripts $ vim ../third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
+# remove '~' from arm64 and amd64
+~/trunk/src/scripts $ emerge-amd64-usr --newuse <group>/<package>
+# optional - add missing dependencies, see line 2 ff. above
+~/trunk/src/scripts $ ./build_image
+~/trunk/src/scripts $ ./image_to_vm.sh --from=../build/images/amd64-usr/latest --format qemu
+~/trunk/src/scripts $ ../build/images/amd64-usr/latest/flatcar_production_qemu.sh &
```
pothos, line 350: The qemu scripts may write to the terminal, I would not use `&` here but rather recommend the user to open a new terminal tab for SSH outside of this SDK container.

pothos: Actually I would do both outside the SDK container since it's easier to pass an Ignition config to the qemu script this way.

docs/reference/developer-guides/sdk-building-production-images.md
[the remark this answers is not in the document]
t-lo: I'll at least add a note to the "Modifying OS images" doc so people are aware of the issue and know what to do.

t-lo:
Addressed @pothos ' feedback: call out the role of the versionfile and how to handle it, and start the qemu VM on the host, not inside of the container.

docs/reference/developer-guides/sdk-modifying-flatcar.md
pothos: With https://github.com/flatcar-linux/scripts/pull/209 I think I should be happy with the reached state and we can resolve this, maybe it's good to tell people how to get the latest nightly `version.txt` from a tag with `git show REF:sdk_container/…/version.txt` until we push it each night to the maintenance/main branch

docs/reference/developer-guides/sdk-building-production-images.md
```diff
@@ -5,68 +5,140 @@ aliases:
...
+3. creating one or more vendor-specific image files from the generic OS image.
+
+
+Optionally, the build process may include building the SDK from scratch based on a previous - existing - SDK, e.g. to update core build tools and utilities.
+In that case, the above 3 steps are preceded by
+
+1. Compile all core and SDK packages from source to generate a new SDK root FS and build a tarball from that; generate a new SDK release version and set the OS release to the same (SDK) version.
+2. Build a base SDK container image using the tarball from 1.
+   a. build amd64 and arm64 toolchains and related board support
+   b. then, from the image from 2.a, generate from scratch 3 container images - "all", "amd64", and "arm64" with the respective board support included.
```
krnowak, line 99: Is there somewhere a rendering of this document available? Because apparently github markdown does not support `a.` or `a)`, only `1.` or `1)`. And the nested ordered list uses roman numerals. So:
```suggestion
   1. build amd64 and arm64 toolchains and related board support
   2. then, from the image from 2. i., generate from scratch 3 container images - "all", "amd64", and "arm64" with the respective board support included.
```

docs/reference/developer-guides/sdk-tips-and-tricks.md
```diff
@@ -102,51 +96,14 @@ References:
 
 Your SSH keys should be detected and added automatically by the image build process. Optionally, you can set a password for the `core` user which you can use later for ssh authentication, should SSH pubkey authentication not work for you.
 
-After entering the chroot via `cork` for the first time, you can set user `core`'s password:
+After entering the SDK container for the first time (or after re-creating it , you can set user `core`'s password:
```
krnowak, line 99:
```suggestion
After entering the SDK container for the first time (or after re-creating it), you can set user `core`'s password:
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -298,48 +278,134 @@ You can log in via SSH keys or with a different ssh port by running this example
...
+Multiple package sources - in separate directories - can be stacked on top of each other.
+These “overlays” allow custom extensions or even custom sub-trees on top of an existing foundation.
+In these stacks, “upper” level packages override “lower” level ones.
+The Flatcar build system uses a fork of Gentoo upstream’s [portage-stable](https://github.com/flatcar-linux/portage-stable) as its base, and the overlay repository [coreos-overlay](https://github.com/flatcar-linux/coreos-overlay) for Flatcar specific modifications and packages on top.
+
+Packages are built using "ebuild" files.
+These files contain dependencies of a package - both build and runtime - as well as implement callbacks for downloading, patching, building, and installing the package.
+The callbacks in these ebuild files are written in shell.
+The Gentoo package system - portage - will, when building / installing a package, run the respective callbacks in order (e.g. `src_fetch()` for downloading package sources, and `src_compile()` for building).
+Common ebuild functions shared across many packages are implemented via classes (in `eclass/`) which can be inherited by package ebuilds.
```
krnowak, line 307:
```suggestion
Common ebuild functions shared across many packages are implemented via eclasses (in `eclass/`) which can be inherited by package ebuilds.
```

docs/reference/developer-guides/sdk-modifying-flatcar.md
```diff
@@ -348,19 +414,9 @@ Lastly, the SDK might lack unmasks if the respective architecture is masked in t
...
-
-**NOTE**: a bug in the Chromium part of the SDK will lead to the following warning:
-```shell
---- Invalid atom in /build/amd64-usr/etc/portage/package.unmask/cros-workon:
-     =<group>/<package>
-```
-The SDK creates softlinks for both `/build/amd64-usr/etc/portage/package.unmask/cros-workon` and `/build/amd64-usr/etc/portage/package.keywords/cros-workon` to the same file - to `.config/cros_workon/amd64-usr` (outside the SDK chroot). This effectively merges all required keywords and unmasks into the same file. The keywords and unmasks files follow different semantics, leading to emerge printing the warning above. Since invalid atoms will be ignored, it’s safe to ignore the warning for the time being.
+To proceed, add the package name and version, and its masked architectures to the `package.accept_keywords` file inside the `coreos` profile. Which `package.accept_keywords` file should be updated depends on couple factors - whether it is needed for both SDK and OS image or only for SDK or only for OS image, whether it is needed for both AMD64 and ARM64 images, or only for AMD64 or only for ARM64. Please refer to `README.md` in coreos-overlay for a summary about profiles
```
krnowak, line 418:
```suggestion
To proceed, add the package name and version, and its masked architectures to the `package.accept_keywords` file inside the `coreos` profile. Which `package.accept_keywords` file should be updated depends on couple factors - whether it is needed for both SDK and OS image or only for SDK or only for OS image, whether it is needed for both AMD64 and ARM64 images, or only for AMD64 or only for ARM64. Please refer to `README.md` in coreos-overlay for a summary about profiles.
```

krnowak (changes requested):

docs/reference/developer-guides/_index.md
```diff
@@ -5,10 +5,12 @@ aliases:
...
 ---
 
-Most users will never have to build Flatcar Container Linux from source or modify it in any way. If you have a need to modify Flatcar Container Linux, we provide an SDK that allows you to build your own developer images. We also provide OEM functionality for cloud providers and other companies that must customize Flatcar Container Linux to run within their environment.
+This section is aimed at curious developers interested in building Flatcar Container Linux from source and/or in modifying the OS.
+We provide a containerised SDK that allows you to extend Flatcar and to build your own OS images.
+We also provide OEM functionality for cloud providers and similar use cases to customize Flatcar Container Linux to run within their environment.
 
 * [Guide to building custom Flatcar images from source][mod-cl]
-* [Building production images][production-images]
+* [Vending production images / CI integration][production-images]
```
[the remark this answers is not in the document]
t-lo: :rofl:

docs/reference/developer-guides/sdk-modifying-flatcar.md
t-lo: Good point, I'll make a mental note to update the SDK docs with instructions how to use the nightly tags after flatcar-linux/scripts#209 is merged and our CI is updated.

t-lo:
Addressed feedback / added suggestions. Thank you @pothos and @krnowak for all your great help!

pothos (approved):

docs/reference/developer-guides/_index.md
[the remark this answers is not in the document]
krnowak: Hm, is actually "Vending" desired here? I thought it should remain as "Building".

PR #198, opened 2022-01-10
Add more info on the update.conf file and manual updates

pothos:
- setup/releases/update-conf: add more info on how to use the parameters
- setup/releases/update-strategies: document the flatcar-update tool    
    When automatic updates are disabled and the update should be done
    manually it helps to still use the flatcar-update tool as it handles
    the disabling of updates again at the right time.
    Document it as alternative to doing all steps manually.

### DIFF
--- a/docs/setup/releases/update-conf.md
+++ b/docs/setup/releases/update-conf.md
@@ -37,11 +37,12 @@ Default installs of Flatcar will likely not need custom settings, and an empty o
 
 * `GROUP`
   * The channel/group this host will pull updates from
-  * public channels will be: `stable`, `beta`, `alpha`
-  * otherwise these are UUIDs
+  * public channels will be: `stable`, `beta`, `alpha` - since this value is also part of the OS image under `/usr/share/flatcar/update.conf` you should only overwrite it if needed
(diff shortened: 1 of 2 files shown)


tormath1 (approved):
The documentation is good and clear - I just have one question regarding `flatcar-update` but it's not blocking this PR:

> The `flatcar-update` tool automatically removes the `SERVER=disabled` line to apply a manual update and also has a `--disable-afterwards` switch to set `SERVER=disabled` again after applying the update:

To me, `flatcar-update` should preserve the current parameters: if `SERVER=disabled` is set, it should automatically set back this value at the end of the execution. It should not be required to add the `--disable-afterwards`, this is something we can easily forget.

pothos:
> To me, `flatcar-update` should preserve the current parameters: if `SERVER=disabled` is set, it should automatically set back this value at the end of the execution. It should not be required to add the `--disable-afterwards`, this is something we can easily forget.

Just checked, this should already be the case because the previous SERVER value gets set again.

pothos:
Corrected that section

tormath1 (approved):

PR #200, opened 2022-01-17
Update virtualbox.md

alphafalcon:
Raise RAM recommendation to 2 GB - 1GB leads to crashes see https://github.com/coreos/bugs/issues/2455#issuecomment-468224637

# [Update RAM recommendation in virtualbox.md]

Using 1 GB RAM leads to crashes during boot.


jepio:
Thanks for this, indeed 2GB RAM is the minimum that has a chance of working.

PR #207, opened 2022-02-01
Update qemu.md

bobhenkel:
Fixed the incorrect word.

# [Title: Fixed wrong word]
Just updated the wrong word in the sentence.
## How to use
NA
## Testing done
Quick read


pothos (changes requested):
I think this was meant to say that the network configuration can be done differently than done in this guide

PR #211, opened 2022-02-17
docs/setup/releases/update-strategies: document airgapped updates

pothos:
With the flatcar-update tool we finally have a way for airgapped
updates.

## Testing done

None, just adapted the fallback instructions from https://www.flatcar.org/docs/latest/migrating-from-coreos/update-from-container-linux/#the-migration-script to use SSL


docs/setup/releases/update-strategies.md
```diff
@@ -202,6 +202,31 @@ $ sudo cgpt show "$(rootdev -s /usr)" | grep successful=1
                                   Attr: priority=1 tries=0 successful=1
 ```
 
+## Airgapped updates
+
+Updating a machine without Internet access is done in two steps.
+First the wanted update payload is downloaded from a non-airgapped machine, then you copy it to your airgapped machine and run the `flatcar-update` tool.
+If the `flatcar-update` tool is missing on your machine, [download](https://raw.githubusercontent.com/flatcar-linux/flatcar-docs/main/update-to-flatcar.sh) it first, too.
```
krnowak, line 209: Is this a correct link? It seems to point to the script that updates coreos to flatcar, not the flatcar update one.

docs/setup/releases/update-strategies.md
```diff
@@ -202,6 +202,31 @@ $ sudo cgpt show "$(rootdev -s /usr)" | grep successful=1
                                   Attr: priority=1 tries=0 successful=1
 ```
 
+## Airgapped updates
+
+Updating a machine without Internet access is done in two steps.
+First the wanted update payload is downloaded from a non-airgapped machine, then you copy it to your airgapped machine and run the `flatcar-update` tool.
```
krnowak, line 208:
```suggestion
First, you need to download the update payload on a non-airgapped machine, then you copy it to your airgapped machine and run the `flatcar-update` tool.
```

docs/setup/releases/update-strategies.md
pothos: Oh, right

krnowak (approved):

PR #214, opened 2022-03-01
Update switching-to-unified-cgroups.md

jepio:
Add description of new method of switching to cgroupsv1.

### DIFF
--- a/docs/container-runtimes/switching-to-unified-cgroups.md
+++ b/docs/container-runtimes/switching-to-unified-cgroups.md
@@ -63,6 +63,22 @@ systemd:
 
 However, the kernel commandline setting doesn't take effect on the first boot, and a reboot is required before the snippet becomes active.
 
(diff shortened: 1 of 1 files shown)


tormath1 (approved):

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -63,6 +63,22 @@ systemd:
 
 However, the kernel commandline setting doesn't take effect on the first boot, and a reboot is required before the snippet becomes active.
 
+If your deployment can't tolerate the required reboot, consider using the following snippet to switch to legacy cgroups without a reboot. This is supported by Flatcar v3139.1.0 or newer:
```
pothos, line 66:
```suggestion
If your deployment can't tolerate the required reboot, consider using the following snippet to switch to legacy cgroups without a reboot. This is supported by Flatcar 3139.1.0 or newer:
```
We normally don't use the v prefix for versions, I would leave it out for consistency.

pothos (approved):

PR #217, opened 2022-03-04
installing: Rework Overview to be hand-on with QEMU and CLC

pothos:
The existing docs were confusing and not really actionable to give an
interactive introduction experience.
Rework the docs to be hand-on with a local QEMU VM and a small CLC
config that runs an NGINX container.


t-lo (approved):
LGTM, thank you!

PR #219, opened 2022-03-04
docs/installing/_index.md: quickstart intro

t-lo:
This change adds an intro section to the quickstart document, outlining basic concepts of configuring and provisioning Flatcar.

PR #220, opened 2022-03-04
Update ct container references to use ghcr

t-lo:
We moved our ct container to ghcr.io - this change updates the documentation accordingly (except for `docs/installing/_index.md` which is updated in https://github.com/flatcar-linux/flatcar-docs/pull/219).


pothos (approved):
Thanks

PR #223, opened 2022-03-04
Remove references to Edge

pothos:
The Edge channel is not maintained anymore.


tormath1 (approved):
looks good  - I don't understand why git diff is messing up with file like docs/installing/bare-metal/booting-with-pxe.md but seen the same thing on a previous PR

PR #230, opened 2022-03-13
docs/installing/_index: Fix flatcar_production_qemu.sh URL

nielsslot:
# Fix URL to flatcar_production_qemu.sh

Fix the URL and filename of the flatcar_production_qemu.sh file. The incorrect URL has a _image postfix and doesn't exist on the server. The guide later uses the correct filename for the script, which led me to think that the download instructions were wrong.

## Testing done

Did a wget with the correct URL.


pothos (approved):
Thanks!

PR #237, opened 2022-03-25
azure.md: config snippet for using PTP RTC

t-lo:
This change adds a config snippet to the Azure documentation for using a minimal chrony container instead of systemd-timesyncd. Chrony is set up to use Azure's virtual RTC.

Closes https://github.com/flatcar-linux/Flatcar/issues/686 .


docs/installing/cloud/azure.md
```diff
@@ -221,6 +221,75 @@ Transpile it to Ignition JSON:
...
+      contents: |
+        [Unit]
+        Description=Build the chrony container image
+        ConditionPathExists=!/opt/chrony-build/done
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        WorkingDirectory=/opt/chrony-build
+        ExecStart=docker build -t chrony .
```
pothos, line 265: For some reason, sometimes, using a relative path fails and an absolute path was required - not sure if that was an issue of the past or may happen again

t-lo: Good catch, will update.

t-lo: Addressed, thanks for catching this.

docs/installing/cloud/azure.md
```diff
@@ -221,6 +221,75 @@ Transpile it to Ignition JSON:
...
+        [Unit]
+        Description=Build the chrony container image
+        ConditionPathExists=!/opt/chrony-build/done
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        WorkingDirectory=/opt/chrony-build
+        ExecStart=/usr/bin/docker build -t chrony .
+        ExecStartPost=touch done
```
pothos, line 266:
```suggestion
        ExecStartPost=/usr/bin/touch done
```

pothos (approved):

PR #241, opened 2022-04-14
Update create_cgroupv1_ami.sh

jepio:
AWS only accepts stream optimized VMDKs, so the final `qemu-img convert` from raw to vmdk needs to
be passed `-o subformat=streamOptimized` so that the vmdk is accepted. There have been reports
of stream vmdk support in qemu-img being buggy, but at least one user report confirms that with this
fix the produced VMDK works (and verity would detect most silent corruption to the image).


pothos (approved):
thanks

PR #248, opened 2022-05-27
Add Ignition v3 kargs docs for controlling autologin

pothos:
With Ignition v3 it became easier to control the kernel parameters in
grub.cfg. This helps users to enable/disable autologin as wanted
without doing a manual reboot.
Document how to use the Ignition v3 kernelArguments directive through
with the Butane Flatcar variant to enable/disable autologin.

## Testing done

Transpiled the Butane config to Ignition, didn't test the Equinix Metal CLC for Ignition v2


docs/setup/customization/other-settings.md
```diff
@@ -89,9 +89,42 @@ Further details can be found in the systemd man pages:
...
+      "quiet"
+    ]
+  }
+}
+```
+
+Instead of using `kernelArguments` you can also use the plain file directive in Ignition to write to `/usr/share/oem/grub.cfg`.
+However, because Ignition runs after GRUB, the GRUB configuration won't take effect until the next reboot of the node.
+
+Here's an example Container Linux Cnfiguration for using the plain file directive (this YAML content has to be transpiled to Ignition JSON with `ct`):
```
tormath1, line 127:
```suggestion
Here's an example Container Linux Configuration for using the plain file directive (this YAML content has to be transpiled to Ignition JSON with `ct`):
```

tormath1 (approved):

PR #252, opened 2022-07-08
Document how to avoid mismatching mounts after reboot

pothos:
The kernel assigns /dev/sdX nodes by the order the disk is discovered
which can be different each boot. When a path such as /dev/sda is used
for mounting, this can result in a mismatch after reboot.
The better way is to always match the disk by the content that is
expected through the paths under /dev/disk/by-label/ and make sure the
labels get created on the first boot. An alternative when this is not
possible is to use /dev/disk/by-path/.

### DIFF
--- a/docs/setup/storage/mounting-storage.md
+++ b/docs/setup/storage/mounting-storage.md
@@ -11,6 +11,8 @@ Container Linux Configs can be used to format and attach additional filesystems 
 
 Mount units name the source filesystem and target mount point, and optionally the filesystem type. *Systemd* mounts filesystems defined in such units at boot time. The following example formats an [EC2 ephemeral disk][ec2-disk] and then mounts it at the node's `/media/ephemeral` directory. The mount unit is therefore named `media-ephemeral.mount`.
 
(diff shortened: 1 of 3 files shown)


docs/setup/storage/mounting-storage.md
```diff
@@ -11,6 +11,8 @@ Container Linux Configs can be used to format and attach additional filesystems
...
 
+Note that you should not use the direct path `/dev/sdX`for the `What=` path but **use a stable identifier** such as `/dev/disk/by-label/X` or `/dev/disk/by-partlabel/X` because, e.g., `/dev/sda` can become `/dev/sdb` after reboot as the Linux kernel assigns the devices in the order they appear which can be unstable. The best idea is to match the disk based on the content you expect, such as a filesystem or partition label that you set up through formatting the disk on first boot via Ignition's `mount:` directive. This way you can use `/dev/sdX` as the `mount: device:` path which is only used on the first boot and don't have to care whether it will get different names after reboot because your mount unit uses `/dev/disk/by-label/` to find the correct disk. If that is not possible you can try your luck with `/dev/disk/by-path/X` entries that depend on the way the disk are attached to the machine but not on the discovery order of the Linux kernel.
```
dongsupark, line 14:
```suggestion
Note that you should not use the direct path `/dev/sdX`for the `What=` path but **use a distinct identifier** such as `/dev/disk/by-label/X` or `/dev/disk/by-partlabel/X` because, e.g., `/dev/sda` can become `/dev/sdb` after reboot as the Linux kernel assigns the devices in the order they appear which can be unstable. The best idea is to match the disk based on the content you expect, such as a filesystem or partition label that you set up through formatting the disk on first boot via Ignition's `mount:` directive. This way you can use `/dev/sdX` as the `mount: device:` path which is only used on the first boot and don't have to care whether it will get different names after reboot because your mount unit uses `/dev/disk/by-label/` to find the correct disk. If that is not possible you can try your luck with `/dev/disk/by-path/X` entries that depend on the way the disk are attached to the machine but not on the discovery order of the Linux kernel.
```

pothos: Does `distinct stable identifier` work? The stable refers to the behavior that it doesn't change on reboot when the discovery ordering is different.

dongsupark: Yes

dongsupark (approved):

PR #254, opened 2022-08-17
hardening-guide: Document how to disable the core user

pothos:
Since the group gets provided by nss-altfiles on runtime from the
read-only user DB in /usr, the groups docker and wheel can't be changed
to remove the core user.
The solution to restrict access is to disable the core user, restrict
sudo access, or to
change the permissions of the Docker socket.
The rkt section can be deleted as rkt is gone.

### DIFF
--- a/docs/setup/security/hardening-guide.md
+++ b/docs/setup/security/hardening-guide.md
@@ -82,6 +82,20 @@ Recent Intel CPU vulnerabilities cannot be fully mitigated in software without d
 
 The [SMT on Container Linux guide][smt-guide] provides guidance and instructions for disabling SMT.
 
+### Disable USB
+
+If you don't expect to ever use USB, you can disable the kernel module, here a CLC snippet:
+
+```
+storage:
+  files:
+    - path: /etc/modprobe.d/blacklist.conf
+      mode: 0644
+      contents:
+        inline: |
+          blacklist usb-storage
+```
+
 ### SELinux
 
 SELinux is a fine-grained access control mechanism integrated into Flatcar Container Linux. Each container runs in its own independent SELinux context, increasing isolation between containers and providing another layer of protection should a container be compromised.
--- a/docs/setup/security/hardening-guide.md
+++ b/docs/setup/security/hardening-guide.md
@@ -34,15 +34,45 @@ etcd and Locksmith should be secured and authenticated using TLS if you are usin
 
 ### Local users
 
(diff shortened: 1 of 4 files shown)


docs/setup/security/hardening-guide.md
```diff
@@ -34,15 +34,27 @@ etcd and Locksmith should be secured and authenticated using TLS if you are usin
...
+The docker daemon is accessible via a unix domain socket at `/run/docker.sock`. Users in the "docker" group have access to this service and access to the docker socket grants similar capabilities to sudo. The core user, by default, has access to the docker group. The group can't be easily changed and thus the solution to restrict access is to disable login for the `core` user or restrict the Docker socket permissions.
 
-### rkt fetch
+You can restrict the Docker socket to root by creating a unit drop-in for `docker.socket` in `/etc/systemd/system/docker.socket.d/10-restrict.conf`:
 
-Users in the "rkt" group have access to the rkt container image store. A user may download new images and place them in the store if they belong to this group. This could be used as an attack vector to insert images that are later executed as root by the rkt container runtime. The core user, by default, has access to the rkt group. You can change this by removing the core user from rkt by running this command: `gpasswd -d core rkt`.
+```
+[Socket]
+SocketGroup=root
+```
```
krnowak, line 57: Probably having this example as a CLC snippet would be a more useful, copy-paste-ready example?

docs/setup/security/hardening-guide.md
```diff
@@ -34,15 +34,35 @@ etcd and Locksmith should be secured and authenticated using TLS if you are usin
 
 ### Local users
 
-Flatcar Container Linux has a single default user account called "core". Generally this user is the one that gets ssh keys added to it via a Container Linux Config for administrators to login. The core user, by default, has access to the wheel group which grants sudo access. You can change this by removing the core user from wheel by running this command: `gpasswd -d core wheel`.
+Flatcar Container Linux has a single default user account called "core". Generally this user is the one that gets ssh keys added to it via a Container Linux Config for administrators to login. The core user, by default, has access to the wheel group which grants sudo access. The group can't be easily changed and thus the solution to restrict access is to either require a password for sudo but not setting one, or disable login for the `core` user.
+
+A sudo drop-in can be created under `/etc/sudoers.d/core-passwd` with the contents `core	ALL=(ALL) 	ALL` and as long as the core user has no password set it can't use `sudo`.
```
krnowak, line 39: How about (just drop those backslashes there):

```suggestion
A sudo drop-in can be created under `/etc/sudoers.d/core-passwd` with the contents `core	ALL=(ALL) 	ALL` and as long as the core user has no password set it can't use `sudo`. Here is a CLC snippet:

\```
storage:
  files:
    - path: /etc/sudoers.d/core-passwd
      mode: 0644
      contents:
        inline: |
          core	ALL=(ALL) 	ALL
\```
```

PR #258, opened 2022-09-07
Prepare github org rename

pothos:
The "flatcar-linux" org will be renamed to "flatcar". For renames, there are no redirections in place and we have to update all links.

## How to use

Merge when rename is done


pothos (commented):
We also have to rename the ghcr.io/flatcar-linux references to ghcr.io/flatcar - note that there are no redirects!

pothos:
Will merge soon

dongsupark (approved):

PR #260, opened 2022-09-22
Initial update of docs to utilize Butane configs

tylerauerbeck:
# Changing examples to use Butane Configs

This PR will update the majority of docs to utilize Butane configs instead of CLC configs

Begins to address https://github.com/flatcar/Flatcar/issues/852

## How to use

Users should be able to look at the updated docs and be able to successfully use Butane to render Ignition files.

## Testing done

Have run each updated Butane config to ensure equivalent output is rendered.


pothos:
Thanks a lot!

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -34,23 +34,21 @@ reboot
...
-      path: "/grub.cfg"
+    - path: "/grub.cfg"
       mode: 0644
-      append: true
-      contents:
-        inline: |
-          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller"
+      append:
+        - inline: |
+            set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller"
```
pothos, line 51: We could keep this and the below for reference use with LTS-2022 (3033.x.y). Then I would migrate away from `append: inline:` on `grub.cfg` and instead of the Ignition v3 kernel args setting. Since it handles the reboot we just need the kargs example with Butane now and the existing two configs can stay CLC with the information that they should be used on LTS which doesn't support Ignition v3 and the kargs setting.

docs/installing/_index.md
```diff
@@ -17,33 +17,33 @@ We'll cover
 
 Flatcar Container Linux is configured at _provisioning time_.
 There are two configuration languages to set up Flatcar, aimed at different use cases:
-- [Container Linux Config][cl-configs] (CLC) is human-readable / writable YAML and must be converted (transpiled) into Ignition config before Flatcar can use it.
+- [Container Linux Config][butane-configs] Butane is human-readable / writable YAML and must be converted (transpiled) into Ignition config before Flatcar can use it.
```
pothos, line 20:
```suggestion
- [Butane][butane-configs] is human-readable / writable YAML and must be converted (transpiled) into Ignition config before Flatcar can use it. It's the successor of [Container Linux Config][cl-configs] which is also still supported (Butane is not supported for LTS-2022).
```

docs/installing/_index.md
```diff
@@ -89,29 +91,16 @@ systemd:
...
-    "units": [
-      {
-        "contents": "[Unit]\nDescription=NGINX example\nAfter=docker.service\nRequires=docker.service\n[Service]\nTimeoutStartSec=0\nExecStartPre=-/usr/bin/docker rm --force nginx1\nExecStart=/usr/bin/docker run --name nginx1 --pull always --net host docker.io/nginx:1\nExecStop=/usr/bin/docker stop busybox1\nRestart=always\nRestartSec=5s\n[Install]\nWantedBy=multi-user.target\n",
-        "enabled": true,
-        "name": "nginx.service"
-      }
-    ]
-  }
-}
+{"ignition":{"version":"3.3.0"},"systemd":{"units":[{"contents":"[Unit]\nDescription=NGINX example\nAfter=docker.service\nRequires=docker.service\n[Service]\nTimeoutStartSec=0\nExecStartPre=-/usr/bin/docker rm --force nginx1\nExecStart=/usr/bin/docker run --name nginx1 --pull always --net host docker.io/nginx:1\nExecStop=/usr/bin/docker stop nginx1\nRestart=always\nRestartSec=5s\n[Install]\nWantedBy=multi-user.target\n","enabled":true,"name":"nginx.service"}]}}
```
pothos, line 103: Would be good to pipe this through `jq` once for formatting

docs/installing/cloud/equinix-metal.md
```diff
@@ -106,43 +108,6 @@ kernel_arguments:
...
-systemd:
-  units:
-    - name: getty@.service
-      dropins:
-        - name: 10-autologin.conf
-          contents: |
-            [Service]
-            ExecStart=
-            ExecStart=-/sbin/agetty --noclear %I $TERM
-```
```
pothos, line 143: Let's keep this around, or?

pothos: It's an alternative to setting the kernel args and saves a reboot

docs/installing/cloud/vmware.md
```diff
@@ -155,11 +156,13 @@ IP configuration specified via `guestinfo.interface.*` and `guestinfo.dns.*` var
...
+If you want to use dynamic data such as `{PRIVATE_IPV4}` and `{PUBLIC_IPV4}` in your Butane Config, you have to use the `--platform=custom` argument to the config transpiler and define your own `coreos-metadata.service`.
 
-In the following example we will use the [reserved variables](https://github.com/kinvolk/afterburn/blob/master/docs/container-linux-legacy.md) `COREOS_CUSTOM_PUBLIC_IPV4` and `COREOS_CUSTOM_PRIVATE_IPV4` known to the config transpiler so that Container Linux Configs which contain `{PUBLIC_IPV4}` in a systemd unit will use `${COREOS_CUSTOM_PUBLIC_IPV4}` instead by automatically sourcing it via `EnvironmentFile=/run/metadata/coreos`.
+In the following example we will use the [reserved variables](https://github.com/kinvolk/afterburn/blob/master/docs/container-linux-legacy.md) `COREOS_CUSTOM_PUBLIC_IPV4` and `COREOS_CUSTOM_PRIVATE_IPV4` known to the config transpiler so that Butane Configs which contain `{PUBLIC_IPV4}` in a systemd unit will use `${COREOS_CUSTOM_PUBLIC_IPV4}` instead by automatically sourcing it via `EnvironmentFile=/run/metadata/coreos`.
```
pothos, line 161: We need to rework this. With Butane we don't support the `{PUBLIC_IPV4}` format but instead it must be done manually by doing what the resulting ignition service declaration does.

docs/installing/cloud/vmware.md
```diff
@@ -176,10 +179,6 @@ systemd:
         Environment=OUTPUT=/run/metadata/coreos
         ExecStart=/usr/bin/mkdir --parent /run/metadata
         ExecStart=/usr/bin/bash -c 'echo "COREOS_CUSTOM_PRIVATE_IPV4=$(ip addr show ens192 | grep "inet 10." | grep -Po "inet \K[\d.]+")\nCOREOS_CUSTOM_PUBLIC_IPV4=$(ip addr show ens192 | grep -v "inet 10." | grep -Po "inet \K[\d.]+")" > ${OUTPUT}'
-
-etcd:
-  # Now we can use dynamic data with {PRIVATE_IPV4} and {PUBLIC_IPV4}
-  advertise_client_urls:       "http://{PUBLIC_IPV4}:2379"
```
pothos, line 182: Would be good to put an updated example here, maybe with the resulting etcd service file

docs/installing/cloud/vmware.md
```diff
@@ -297,48 +296,11 @@ kernel_arguments:
...
-    - path: /grub.cfg
-      filesystem: oem
-      mode: 0644
-      contents:
-        inline: |
-          set oem_id="vmware"
-          set linux_append=""
-```
-
-To take effect directly on first boot, the alternative is to create a `getty@.service` drop-in, here a CLC snippet:
```
pothos, line 322: same as for EM

docs/installing/community-platforms/openstack.md
```diff
@@ -116,10 +118,11 @@ systemd:
 Transpile it to Ignition JSON:
 
 ```shell
-cat cl.yaml | docker run --rm -i ghcr.io/flatcar/ct:latest -platform openstack-metadata > ignition.json
+cat cl.yaml | docker run --rm -i quay.io/coreos/butane:latest > ignition.json
 ```
 
-The `{PRIVATE_IPV4}` and `{PUBLIC_IPV4}` substitution variables are fully supported in Container Linux Configs on OpenStack deployments using the metadata service. Unfortunately systems relying on config drive are currently unsupported.
+<!--- TODO: Do the below substitution variables actually work with butane? -->
```
pothos, line 124: No, this was syntactic sugar and needs to be done now directly in the systemd service, using the metadata file and the right variable name as the resulting tranpiled ignition config did

pothos (commented):
The rename of the folder from cl-config/ to butane-config/ would break URLs and at the moment the index/examples/from-cloud-config docs are not updated. I would keep the folder name as is for now.

tylerauerbeck:
@pothos Correct, I haven't made it all the way through the examples yet, so not everything has been updated yet. Just wanted to drop my process here so folks know that this was being worked on and to see if I was heading down the right path. Would we still want to keep `cl-config` after everything has been updated?

pothos:
I suggest we migrate https://www.flatcar.org/docs/latest/provisioning/config-transpiler/ to use butane and move the essential parts of the `cl-config` section there but we keep https://www.flatcar.org/docs/latest/provisioning/cl-config/ as is for LTS and maybe move the CLC spec there.

docs/installing/cloud/vmware.md
tylerauerbeck: Going to rework the VMware doc.

docs/installing/cloud/vmware.md
tylerauerbeck: So since `etcd` isn't available in `butane` , you're thinking about just crafting the service files through `storage.files`?

docs/installing/community-platforms/openstack.md
tylerauerbeck: Will be reworking the openstack doc as well.

tylerauerbeck:
@pothos Reworked some of the comments and issues you mentioned above. Feel free to let me know if you see any issues. I'll be working to cleanup the the `vmware` and `openstack` docs. I'll also look to add similar `configuration` and `examples` docs that exist for `ct` if that makes sense. That way there are roughly equivalent reference docs for ct and butane.

docs/provisioning/config-transpiler/_index.md
```diff
@@ -1,39 +1,80 @@
...
 
-ct on the other hand _does_ understand the specifics of Container Linux. A user can merely specify an etcd version and some etcd options, and ct knows that there's a unit called `etcd-member` already on the system it can enable. It knows what options are supported by etcd, so it can sanity check them for the user. It can then generate an appropriate systemd dropin with the user's options, and provide Ignition the level of verbosity it needs, that would be tedious for a human to create.
+Butane on the other hand _does_ understand the specifics of Container Linux. A user can merely specify an etcd version and some etcd options, and butane knows that there's a unit called `etcd-member` already on the system it can enable. It knows what options are supported by etcd, so it can sanity check them for the user. It can then generate an appropriate systemd dropin with the user's options, and provide Ignition the level of verbosity it needs, that would be tedious for a human to create.
```
tormath1, line 41: I think this not the case with `butane`. It's not aware about etcd-member, locksmith, etc.

docs/provisioning/config-transpiler/_index.md
```diff
@@ -1,39 +1,80 @@
...
+variant: flatcar
+version: 1.0.0
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys:
+        - ssh-rsa AAAAB......xyz email@host.net
+```
+
+To turn this Butane Config into a usable Ignition Config, we can then run: `docker run --rm -i quay.io/coreos/butane:latest`. This will result in the above YAML being turned into the below JSON
```
tormath1, line 57:
```suggestion
To turn this Butane Config into a usable Ignition Config, we can then run: `docker run --rm -i quay.io/coreos/butane:latest < your_config.yaml > your_config.json`. This will result in the above YAML being turned into the below JSON
```

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -66,23 +63,18 @@ However, the kernel commandline setting doesn't take effect on the first boot, a
...
-        inline: |
-          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller"
+    - device: "/dev/disk/by-label/OEM"
+      format: "btrfs"
   files:
     - path: /etc/flatcar-cgroupv1
       mode: 0444
+kernel_arguments:
+  should_exist:
+    - set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller
```
tormath1, line 77: Maybe we can directly use this format:
```suggestion
    - systemd.unified_cgroup_hierarchy=0
    - systemd.legacy_systemd_cgroup_controller
```

What do you think @pothos ?

tormath1 (commented):
LGTM, thanks a lot for this initial contribution. I think we should merge this as a solid foundation to iterate over in follow-up PRs (for adding vmware, openstack for example) and fixing leftovers.

docs/provisioning/config-transpiler/_index.md
pothos: Butane's Flatcar variant is aware of Flatcar specifics, maybe we can talk about this (currently it just means that generating clevis sections is caught because they are not supported)

docs/container-runtimes/switching-to-unified-cgroups.md
pothos: Yes, the `set linux_append=` is wrong here

docs/container-runtimes/switching-to-unified-cgroups.md
```diff
@@ -34,23 +34,20 @@ reboot
...
-    - filesystem: "OEM"
-      path: "/grub.cfg"
-      mode: 0644
-      append: true
-      contents:
-        inline: |
-          set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller"
+kernel_arguments:
+  should_exist:
+    - set linux_append="$linux_append systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller
```
tormath1, line 50: Same as below :) 
```suggestion
    - systemd.unified_cgroup_hierarchy=0
    - systemd.legacy_systemd_cgroup_controller
```

docs/installing/cloud/azure.md
```diff
@@ -501,10 +553,13 @@ For each machine in the list, you should have a `machine-NAME.yaml.tmpl` file wi
...
+version: 1.0.0
 passwd:
   users:
     - name: core
-      ssh_authorized_keys: ${ssh_keys}
+      ssh_authorized_keys: 
+        - ${ssh_keys}
 storage:
   files:
     - path: /home/core/works
```
tormath1, line 565: Can't select, but the `filesystem: root` can be dropped :)

docs/installing/cloud/digitalocean.md
```diff
@@ -347,10 +336,13 @@ For example, create the configuration for `mynode` in the file `machine-mynode.y
...
+version: 1.0.0
 passwd:
   users:
     - name: core
-      ssh_authorized_keys: ${ssh_keys}
+      ssh_authorized_keys: 
+        - ${ssh_keys}
 storage:
   files:
     - path: /home/core/works
```
tormath1, line 348: Same as above: can't select, but the `filesystem: root` can be dropped :)

docs/installing/cloud/equinix-metal.md
```diff
@@ -268,10 +264,13 @@ Create the configuration for `mynode` in the file `machine-mynode.yaml.tmpl`:
...
+version: 1.0.0
 passwd:
   users:
     - name: core
-      ssh_authorized_keys: ${ssh_keys}
+      ssh_authorized_keys: 
+        - ${ssh_keys}
 storage:
   files:
     - path: /home/core/works
```
tormath1, line 276: Same as above: can't select, but the `filesystem: root` can be dropped :)

tormath1:
@tylerauerbeck I did a final review, let's merge this and thanks again for pioneering this task!

docs/provisioning/config-transpiler/_index.md
pothos:
```suggestion
Butane on the other hand _does_ understand the specifics of Flatcar Container Linux. A user currently can't specify Clevis options on Flatcar and Butane does these sanity checks for the user.
```

pothos:
Thanks again, we've picked up your changes in another branch and added some on top and it's now merged.

PR #261, opened 2022-10-14
dev-guides: fix order of kernel package compilation

jepio:
coreos-modules needs to be rebuilt before coreos-kernel, because the kernel ebuild builds the initramfs and this bundles some of the built kernel modules.

The previous steps in the guide, also need an update because they have the same issue but I haven't fully understood how to correct them yet.

### DIFF
--- a/docs/reference/developer-guides/sdk-modifying-flatcar.md
+++ b/docs/reference/developer-guides/sdk-modifying-flatcar.md
@@ -524,8 +524,8 @@ core@localhost ~ $ ...
 ~/trunk/src/scripts $ cd ../third_party/coreos-overlay/sys-kernel/coreos-modules/files/
 ~/trunk/src/scripts $ vim -O commonconfig* amd64_defconfig* my.diff
 ~/trunk/src/scripts $ rm my.diff
-~/trunk/src/scripts $ emerge-amd64-usr sys-kernel/coreos-kernel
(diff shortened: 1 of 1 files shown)


jepio:
@t-lo do you have an idea how to fix the steps that use `ebuild-amd64-usr` directly and operate on the ebuild. A user has had difficulties following the steps: https://github.com/flatcar/Flatcar/issues/871#issuecomment-1278637887

t-lo:
Great catch! I believe your proposed solution is worth a shot. I'm not sure at which point exactly the kernel config is re-generated but emerging the kernel first might use an outdated config and outdated modules (and also affect the initrd as you stated) .
Since we've built (or downloaded) all packages in the preparation step, emerge thinks the kernel modules binpkg available on the system is up to date and will use it (kernel depends on modules). But since we changed the kernel configuration - which is in the modules ebuild - and did not bump the modules ebuild version, the modules binpkg is in fact out of date. Pulling the emerge modules step before emerge kernel might address this.

t-lo:
@jepio I think this looks good and I feel tempted to approve even though the PR is in Draft state :smile_cat:

jepio:
i will still need to look and fix the other portion.

jepio:
@t-lo have another look, I think the updated steps should work.

PR #262, opened 2022-10-24
customization/acpi: add section about qemu-guest-agent

jepio:
Qemu-guest-agent is used in some of the same contexts as acpid, so mention how this works, and the option needed for openstack instances.


pothos (approved):

jepio:
Good idea, will do that next.

tormath1 (approved):

PR #264, opened 2022-10-26
Update libvirt.md

scuot:
Mv command was incomplete 

# Mv command was incomplete

Completed the move command to work correctly.

## How to use

Move the file?

## Testing done

yes


jepio (approved):

jepio:
Thank you @scuot!

PR #267, opened 2022-11-08
fix mismatch in file names

alirezaghey:
Fixed small mismatch between the file name in the config file and its reference in the description below it.


pothos (approved):

pothos:
Thanks

PR #271, opened 2022-11-24
Update link to chromium disk layout

sfudeus:
# Update link to chromium disk layout

The prior chromium site complains that the link is outdated and should be updated.

## How to use

Open the new URL :-)

## Testing done

Checked the new URL to be working and that it show the expected content. 

I suppose the following template items would not be relevant for this kind of change, right?


pothos (approved):

pothos:
Thanks

PR #272, opened 2022-11-27
hetzner: Add link to GitHub repo

pothos:
The link to the repo was missing for Hetzner.

Like with the other Terraform cloud provider docs, link to the repo at the end.

### DIFF
--- a/docs/installing/cloud/hetzner.md
+++ b/docs/installing/cloud/hetzner.md
@@ -213,3 +213,5 @@ terraform apply
 Log in via `ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null core@IPADDRESS` with the printed IP address.
 
 When you make a change to `machine-mynode.yaml.tmpl` and run `terraform apply` again, the machine will be replaced.
+
+You can find this Terraform module in the repository for [Flatcar Terraform examples](https://github.com/flatcar/flatcar-terraform/tree/main/hetzner).
--- a/docs/installing/cloud/hetzner.md
+++ b/docs/installing/cloud/hetzner.md
@@ -39,6 +39,19 @@ Read more about using Terraform and Flatcar [here](../../provisioning/terraform/
 The following Terraform v0.13 module may serve as a base for your own setup.
 It will also take care of registering your SSH key at Hetzner.
 
+You can clone the setup from the [Flatcar Terraform examples repository](https://github.com/flatcar/flatcar-terraform/tree/main/flatcar-terraform-hetzner) or create the files manually as we go through them and explain each one.
+
+```
+git clone https://github.com/flatcar/flatcar-terraform.git
+# From here on you could directly run it, TLDR:
+cd flatcar-terraform-hetzner
+export HCLOUD_TOKEN=...
+terraform init
+# Edit the server configs or just go ahead with the default example
+terraform plan
+terraform apply
+```
+
 Start with a `hetzner-machines.tf` file that contains the main declarations:
 
 ```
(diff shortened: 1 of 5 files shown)


t-lo:
Would it make sense to remove the terraform code from the docs entirely and replace it with something like:

```shell
   git clone https://github.com/flatcar/flatcar-terraform.git
   cd flatcar-terraform-hetzner
   terraform init
   # edit server-configs/server1.yaml or just go ahead with the default NGINX example
   terraform plan
   terraform apply
```

pothos:
> Would it make sense to remove the terraform code from the docs entirely and replace it with something like:

The idea was that people with some existing terraform code could directly copy paste the parts they need from the docs. Maybe there are other ways of linking/embedding the content to make this simple, too.

t-lo:
> > Would it make sense to remove the terraform code from the docs entirely and replace it with something like:
> 
> The idea was that people with some existing terraform code could directly copy paste the parts they need from the docs. Maybe there are other ways of linking/embedding the content to make this simple, too.

Alternatively we could start the page with a brief section discussing the repo (e.g. using the 6 lines I pasted above) and then elaborate after that with a full-blown snippet. Motivation would be to first make provisioning as straightforward as with clouds that officially support Flatcar - unfortunately, people _have to_ use terraform for Hetzner right now as we're not directly supported -  then discuss advanced features like integration into existing automation after that.

pothos:
Thanks, have added this and if ok, will copy it to the other terraform docs.

pothos:
Ok, I would merge it as is and then for https://github.com/flatcar/flatcar-terraform/pull/12 we need to rework the docs to use Butane and match the new setup.

docs/installing/cloud/hetzner.md
```diff
@@ -39,6 +39,19 @@ Read more about using Terraform and Flatcar [here](../../provisioning/terraform/
 The following Terraform v0.13 module may serve as a base for your own setup.
 It will also take care of registering your SSH key at Hetzner.
 
+You can clone the setup from the [Flatcar Terraform examples repository](https://github.com/flatcar/flatcar-terraform/tree/main/hetzner) or create the files manually as we go through them and explain each one.
+
+```
+git clone https://github.com/flatcar/flatcar-terraform.git
+# From here on you could directly run it, TLDR:
+cd hetzner
```
t-lo, line 47:
```suggestion
cd flatcar-terraform-hetzner
```

docs/installing/cloud/hetzner.md
```diff
@@ -39,6 +39,19 @@ Read more about using Terraform and Flatcar [here](../../provisioning/terraform/
 The following Terraform v0.13 module may serve as a base for your own setup.
 It will also take care of registering your SSH key at Hetzner.
 
+You can clone the setup from the [Flatcar Terraform examples repository](https://github.com/flatcar/flatcar-terraform/tree/main/hetzner) or create the files manually as we go through them and explain each one.
```
t-lo, line 42:
```suggestion
You can clone the setup from the [Flatcar Terraform examples repository](https://github.com/flatcar/flatcar-terraform/tree/main/flatcar-terraform-hetzner) or create the files manually as we go through them and explain each one.
```

t-lo (approved):
Added suggestions to update the Hetzner folder in the terraform repo in accordance with https://github.com/flatcar/flatcar-terraform/pull/12 .
LGTM otherwise, thank you!

docs/installing/cloud/hetzner.md
pothos: When we do that change we should also update the listing below to use the new filenames/procedure.

t-lo: Not sure if updating the below is equally urgent; but as soon as https://github.com/flatcar/flatcar-terraform/pull/12 is merged that text above will break, while the details section below would still work (thought be outdated).

PR #276, opened 2023-01-25
sdk-modifying-flatcar: update org name in container registry link

bgilbert:
The old link still works, but points to a page that doesn't have the SDK images.


jepio (approved):
Thanks, that was one of our learnings from the org rename (registries stay and have no redirects).

PR #279, opened 2023-03-10
Document Alpha 3535.0.0 changes for /etc overlay and flatcar-reset

pothos:
- Document flatcar-reset tool
- Document the /etc overlay cleanup logic and how to opt-out

Later I'll also look into updating the Terraform example setup to make use of flatcar-reset.

### DIFF
--- a/docs/provisioning/ignition/_index.md
+++ b/docs/provisioning/ignition/_index.md
@@ -36,7 +36,7 @@ The lack of variable substitution in Ignition has an added benefit of leveling t
 
 ### When is Ignition executed
 
-On boot, GRUB checks the EFI System Partition for a file at `flatcar/first_boot` (or `coreos/first_boot` if the machine was updated from CoreOS CL) and sets `flatcar.first_boot=detected` if found. The `flatcar.first_boot` parameter is processed by a [systemd-generator] in the [initramfs] and if the parameter value is non-zero, the Ignition units are set as dependencies of `initrd.target`, causing Ignition to run. If the parameter is set to the special value `detected`, the `flatcar/first_boot` (or `coreos/first_boot`) file is deleted after Ignition runs successfully.
+On boot, GRUB checks the EFI System Partition for a file at `flatcar/first_boot` (or `coreos/first_boot` if the machine was updated from CoreOS CL) and sets `flatcar.first_boot=detected` if found. The `flatcar.first_boot` parameter is processed by a [systemd-generator] in the [initramfs] and if the parameter value is non-zero, the Ignition units are set as dependencies of `initrd.target`, causing Ignition to run. If the parameter is set to the special value `detected`, the `flatcar/first_boot` (or `coreos/first_boot`) file is deleted after Ignition runs successfully. You can schedule a re-run of Ignition with the `flatcar-reset` tool (available since Alpha 3535.0.0), which also takes care of cleaning up old rootfs state and keeping only the data from the rootfs you want to keep.
 
 Note that [PXE][supported-platforms] deployments don't use GRUB to boot, so `flatcar.first_boot=1` must be added to the boot arguments in order for Ignition to run. `detected` should not be specified so Ignition will not attempt to delete `flatcar/first_boot` (or `coreos/first_boot`).
 
--- a/docs/provisioning/ignition/boot-process.md
+++ b/docs/provisioning/ignition/boot-process.md
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
 
 ### Reprovisioning
 
-To trigger a new Ignition run, either manually set `flatcar.first_boot=1` as temporary kernel command line parameter in GRUB or create the flag file `touch /boot/flatcar/first_boot` (or `/boot/coreos/first_boot` if the machine was updated from CoreOS CL).
+To trigger a new Ignition run, you should use the `flatcar-reset` tool (available from Alpha 3535.0.0) for a (selective) cleanup of the root filesystem during the next boot. It takes care of cleaning up old state (e.g., files from the old configuration or any side effects such as state files) while keeping only the data you want to keep through the `--keep-paths` argument. The paths to keep can be specified as regular expressions. The machine ID can be kept through the `--keep-machine-id` argument (turning it into a kernel cmdline parameter because `/etc/machine-id` can't be preserved directly for systemd first boot semantics). It is also possible to specify that a local or a particular remote Ignition configuration should be used.
 
+When paths to keep are specified, only needed paths should be used and not those set up by the old Ignition config or side effects of it, to really discard the old configuration state. When a path specified is a directory, the contents are preserved as well because `MYPATH/.*` is automatically appended as an additional regular expression for paths to keep.
+To delete the contents of a directory but keep the directory itself, specify it as an equivalent regular expression in the form of `'^/etc/mypath'`, `'/etc/mypath$'`, `'/etc/mypat[h]'`, `'/etc/(mypath)'`, or `'(/etc/mypath)'`.
+
+The used regular expression language is that of `egrep`. Assuming you specified `/etc/mypath`, you can test which paths will be deleted with the following command (note the `-not`):
+
+```sh
+find / /etc -xdev -regextype egrep -not -regex '(/etc/mypath|/etc/mypath/.*)'
+```
+
+You can test which path will be kept with the following command (note the absence of `-not`):
+
+```sh
+find / /etc -xdev -regextype egrep -regex '(/etc/mypath|/etc/mypath/.*)'
+```
+
+Both `/` and `/etc` need to be specified because `/etc` is an overlay mount.
+
+Meaningful examples are:
+
+- `'/etc/ssh/ssh_host_.*'` to preserve SSH host keys
+- `/var/log` to preserve system logs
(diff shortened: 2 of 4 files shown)


docs/provisioning/ignition/_index.md
```diff
@@ -36,7 +36,7 @@ The lack of variable substitution in Ignition has an added benefit of leveling t
...
+On boot, GRUB checks the EFI System Partition for a file at `flatcar/first_boot` (or `coreos/first_boot` if the machine was updated from CoreOS CL) and sets `flatcar.first_boot=detected` if found. The `flatcar.first_boot` parameter is processed by a [systemd-generator] in the [initramfs] and if the parameter value is non-zero, the Ignition units are set as dependencies of `initrd.target`, causing Ignition to run. If the parameter is set to the special value `detected`, the `flatcar/first_boot` (or `coreos/first_boot`) file is deleted after Ignition runs successfully. You can schedule a re-run of Ignition with the `flatcar-reset` tool (available since Alpha 3535.0.0), which also takes care of cleaning up old rootfs state and only keeping the rootfs data you want to keep.
```
krnowak, line 39:
```suggestion
On boot, GRUB checks the EFI System Partition for a file at `flatcar/first_boot` (or `coreos/first_boot` if the machine was updated from CoreOS CL) and sets `flatcar.first_boot=detected` if found. The `flatcar.first_boot` parameter is processed by a [systemd-generator] in the [initramfs] and if the parameter value is non-zero, the Ignition units are set as dependencies of `initrd.target`, causing Ignition to run. If the parameter is set to the special value `detected`, the `flatcar/first_boot` (or `coreos/first_boot`) file is deleted after Ignition runs successfully. You can schedule a re-run of Ignition with the `flatcar-reset` tool (available since Alpha 3535.0.0), which also takes care of cleaning up the old rootfs state and keeping only the data in rootfs you want to keep.
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
 
+When paths to keep are specified, only needed paths should be used and not those set up by the old Ignition config or side effects of it, to really discard the old configuration state. When a path specified is a folder, the contents are preserved as well because `MYPATH/.*` is automatically appended as additonal regular expression for paths to keep.
+To delete the contents of a folder but keep the folder itself, specify it as equivalent regular expression in the form of `'^/etc/mypath'`, `'/etc/mypath$'`, `'/etc/mypat[h]'`, `'/etc/(mypath)'`, or `'(/etc/mypath)'`.
```
krnowak, line 48:
```suggestion
To delete the contents of a directory but keep the directory itself, specify it as an equivalent regular expression in the form of `'^/etc/mypath'`, `'/etc/mypath$'`, `'/etc/mypat[h]'`, `'/etc/(mypath)'`, or `'(/etc/mypath)'`.
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
+
+The used regular expression language is that of `egrep`. Assuming you specified `/etc/mypath`, you can test which paths will be deleted with (note the `-not`):
```
krnowak, line 50:
```suggestion
The used regular expression language is that of `egrep`. Assuming you specified `/etc/mypath`, you can test which paths will be deleted with the following command (note the `-not`):
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
+
+```sh
+find / /etc -xdev -regextype egrep -not -regex '(/etc/mypath|/etc/mypath/.*)'
+```
+
+You can tests which path will be kept with with (note the absence of `-not`):
```
krnowak, line 56:
```suggestion
You can test which path will be kept with the following command (note the absence of `-not`):
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
+
+```sh
+find / /etc -xdev -regextype egrep -regex '(/etc/mypath|/etc/mypath/.*)'
+```
+
+Both `/` and `/etc` have to be specified because `/etc` is an overlay mount.
```
krnowak, line 62:
```suggestion
Both `/` and `/etc` need to be specified because `/etc` is an overlay mount.
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
+
+Meaningful examples are:
+
+- `'/etc/ssh/ssh_host_.*'` to preserve SSH host keys
+- `/var/log` to preserve system logs
+- `/var/lib/docker` and `/var/lib/containerd` to preserve container state and images
+
+Example for selectively resetting the OS with retriggering Ignition while keeping SSH host keys, logs, and machine ID:
```
krnowak, line 70:
```suggestion
An example for selectively resetting the OS with retriggering Ignition while keeping SSH host keys, logs, and machine ID:
```

docs/provisioning/ignition/boot-process.md
```diff
@@ -42,8 +42,41 @@ After all of the tasks in the initramfs complete, the machine pivots into user s
...
+
+```sh
+sudo flatcar-reset --keep-machine-id --keep-paths '/etc/ssh/ssh_host_.*' /var/log
+sudo systemctl reboot
+```
+
+#### Technical Details for Manual Ignition Re-runs
+
+Not recommended but possible is to either manually set `flatcar.first_boot=1` as temporary kernel command line parameter in GRUB or to create the flag file `touch /boot/flatcar/first_boot` (or `/boot/coreos/first_boot` if the machine was updated from CoreOS CL).
```
krnowak, line 79:
```suggestion
Not recommended but possible is to either manually set `flatcar.first_boot=1` as temporary kernel command line parameter in GRUB or to create the flag file with `touch /boot/flatcar/first_boot` (or `/boot/coreos/first_boot` if the machine was updated from CoreOS CL).
```

docs/provisioning/terraform/_index.md
```diff
@@ -59,9 +59,12 @@ When using a template be careful to refer to the Terraform variables via `${vari
 Sometimes you want to take the declarative approach of Terraform but can't accept that nodes are destroyed and recreated for configuration changes.
 This is the case for nodes that have a manual or slow bring-up process, much data that can't be moved easily, or where the IP address should not change.
 
-Ignition can be told to run again through `touch /boot/flatcar/first_boot` but it [won't clean up any old state][boot-process].
-For that you have to reformat the root filesystem with Ignition to ensure that no old state is present.
-Persistent data should be stored on another partition.
+Ignition can be told to run again through `flatcar-reset` (available since Alpha 3535.0.0), which also takes care of cleaning up old rootfs state and only keeping the rootfs data you want to keep.
```
krnowak, line 62:
```suggestion
Ignition can be told to run again through `flatcar-reset` (available since Alpha 3535.0.0), which also takes care of cleaning up the old rootfs state and keeping only the rootfs data you want to keep.
```

docs/setup/releases/update-strategies.md
```diff
@@ -285,6 +285,17 @@ update_engine_client -reset_status
 update_engine_client -check_for_update
 ```
 
+### Management of config files
+
+Since Alpha 3535.0.0 the OS config files under `/etc` are updated through the overlay mount as long as they are not modified.
+On boot any files in `/etc` that are the same as provided by the booted `/usr/share/flatcar/etc` default for the overlay mount on `/etc` are deleted to ensure that future updates of `/usr/share/flatcar/etc` are propagated - to opt out create `/etc/.no-dup-update` in case you want to keep an unmodified config file as is or because you fear that a future Flatcar version may use the same file as you at which point your copy is cleaned up and any other future Flatcar changes would be applied.
```
krnowak, line 291:
```suggestion
On boot any files in `/etc` that are the same as provided by the booted `/usr/share/flatcar/etc` default for the overlay mount on `/etc` are deleted to ensure that future updates of `/usr/share/flatcar/etc` are propagated. To opt out, create `/etc/.no-dup-update` in case you want to keep an unmodified config file as is or because you fear that a future Flatcar version may use the same file as you at which point your copy is cleaned up and any other future Flatcar changes would be applied.
```

This just made me wonder what happens on rollback? The deleted files are not brought back, are they?

pothos: In older Flatcar versions they will be created again by `systemd-tmpfiles`

docs/provisioning/terraform/_index.md
pothos: Thanks! I went with `cleaning up old state` instead of `the old state` to be more general as it's not a single defined state, but anyway, both would work.

docs/setup/releases/update-strategies.md
krnowak: Oh, cool, so we have this case covered. Thanks.

docs/provisioning/terraform/_index.md
krnowak: Yeah, I was not really sure about "the old state". Looks good.

krnowak (approved):

PR #280, opened 2023-03-13
Fully migrate Butane Config section

pothos:
Rename the Config Transpiler section to Butane Config Transpiler to differenciate it from the Container Linux Config section above. Move out the operator-notes.md which is for CLC, and also move the dynamic data docs to CLC and create a new one under the Ignition section without using CLC features (Butane has no "magic" for that which is actually not bad).


pothos:
Note, we still need to go through the docs to migrate the examples away from special `networkd:`/`etcd:`/`flannel:`/`docker:`/`update:`/`locksmith:` sections, see https://github.com/flatcar/Flatcar/issues/968#issuecomment-1465834953

krnowak (approved):

PR #283, opened 2023-03-17
systemd: Document how to add custom shutdown actions

pothos:
As discussed in https://github.com/flatcar/Flatcar/issues/65 the single path that systemd supports for shutdown hooks does not work for read- only images but we can use regular units to run custom shutdown actions. Document one way to do it (there are many other ways how to express this with systemd units).

## Testing done

Tested this by enabling the unit and rebooting/powering off to check that the marker file appears and `bye` is in the journal.


krnowak:
Or create a sysext that adds the desired shutdown action. :)

pothos:
Unfortunately it gets unloaded in the shutdown before the hooks are processed

krnowak:
> Unfortunately it gets unloaded in the shutdown before the hooks are processed

Rats.

krnowak (approved):

PR #287, opened 2023-03-17
Migrate CLC docs to Butane

sayanchowdhury:
- Migrate the CLC example to Butane config
- Fix the link to the ignition documentation.


sayanchowdhury:
Given LTS still supports CLC, wondering if we should add a section for CLC as well or add somewhere with the Butane [examples](https://github.com/flatcar/flatcar-docs/pull/285) rather than having a CLC section in each page. The CLC section if added could be deprecated on LTS-2022 goes EOL

PR #292, opened 2023-03-17
customization: use butane configuration

tormath1:
for the sshd_config I think we want to overwrite it.

### DIFF
--- a/docs/setup/customization/configuring-date-and-timezone.md
+++ b/docs/setup/customization/configuring-date-and-timezone.md
@@ -73,7 +73,7 @@ NTP synchronized: yes
 
 ## Time synchronization
 
-Flatcar Container Linux clusters use NTP to synchronize the clocks of member nodes, and all machines start an NTP client at boot. Flatcar Container Linux versions later than [681.0.0][681.0.0] use [`systemd-timesyncd(8)`][systemd-timesyncd] as the default NTP client. Earlier versions used [`ntpd(8)`][ntp.org]. Use `systemctl` to check which service is running:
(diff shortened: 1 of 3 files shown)


krnowak (approved):
Overwrite sounds good in this case.

docs/setup/customization/configuring-date-and-timezone.md
```diff
@@ -151,15 +152,17 @@ sudo systemctl enable ntpd
...
 systemd:
   units:
     - name: systemd-timesyncd.service
       mask: true
     - name: ntpd.service
-      enable: true
+      enabled: true
 ```
 
 Because `timesyncd` and `ntpd` are mutually exclusive, it's important to `mask` the `systemd-tinesyncd` service. `Systemctl disable` or `stop` alone will not prevent a default service from starting again.
```
krnowak, line 168:
```suggestion
Because `timesyncd` and `ntpd` are mutually exclusive, it's important to `mask` the `systemd-timesyncd` service. `systemctl disable` or `stop` alone will not prevent a default service from starting again.
```

docs/setup/customization/configuring-date-and-timezone.md
krnowak, line 1: This file mentions Flatcar Container Linux versions later than 681.0.0, which means any Flatcar Container Linux, because I think the earliest version of Flatcar was in 1xxx or 2xxx range. And there's even a link to 681.0.0 manifest somewhere in kinvolk org. Could you simplify it? So "Flatcar Container Linux versions later than 681.0.0 use" could become "Flatcar Container Linux uses".

krnowak (approved):

PR #297, opened 2023-03-30
journald: add examples to show user journal logs

tormath1:
Let's provide some documentation for user journal. 

Note: wait for https://github.com/flatcar/coreos-overlay/pull/2553

### DIFF
--- a/docs/setup/debug/reading-the-system-log.md
+++ b/docs/setup/debug/reading-the-system-log.md
@@ -38,6 +38,21 @@ Dec 22 12:32:39 localhost docker[9772]: /usr/sbin/apache2ctl: 87: ulimit: error 
 Dec 22 12:32:39 localhost docker[9772]: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.6 for ServerName
 ```
 
+## Read the user journal from the current user
+
+It might be required to add a user different from `core` user to the `systemd-journal` group to read the user journal. It can be done with this Butane config:
+```yaml
+variant: flatcar
+version: 1.0.0
+passwd:
+  users:
+    - name: flatcar
+      groups:
+        - systemd-journal
+```
+
+Then from `flatcar` login: `journalctl --user`.
+
 ## Read entries since boot
 
 Reading just the entries since the last boot is an easy way to troubleshoot services that are failing to start properly:


docs/provisioning/config-transpiler/examples.md
```diff
@@ -198,6 +204,8 @@ storage:
 
 This example creates a new systemd user unit called `hello.service`, enables it with an explicit symlink (workaround for Ignition) so it will run on boot, and defines the contents to simply echo `"Hello, World!"`.
 
+_Note_: adding a regular user like "flatcar" can be useful if you want to access the journal logs with `journalctl --user --unit hello.service` otherwise you can access logs with `journactl --user-unit hello.service` from default `core` user.
```
pothos, line 207:
```suggestion
_Note_: Adding a regular user like "flatcar" to the `systemd-journal` group can be useful if you want to access the journal logs with `journalctl --user --unit hello.service`. You can already access logs with `journactl --user-unit hello.service` from the default `core` user.
```

docs/setup/customization/adding-users.md
```diff
@@ -26,6 +28,27 @@ passwd:
       groups: [ sudo, docker ]
 ```
 
+Alternatively, you can use [systemd-userdb](https://www.freedesktop.org/software/systemd/man/systemd-userdbd.service.html) to define membership. Here's the same example with userdb:
```
pothos, line 31:
```suggestion
Because `usermod` does not work to add a user to a predefined system group, you can use [systemd-userdb](https://www.freedesktop.org/software/systemd/man/systemd-userdbd.service.html) to define membership. Here's the same example with userdb:
```

docs/setup/debug/reading-the-system-log.md
```diff
@@ -38,6 +38,24 @@ Dec 22 12:32:39 localhost docker[9772]: /usr/sbin/apache2ctl: 87: ulimit: error
 Dec 22 12:32:39 localhost docker[9772]: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.6 for ServerName
 ```
 
+## Read the user journal from the current user
+
+It might be required to provide a user different from `core` user and in the `systemd-journal` group to read user journal. It can be done with Butane config:
```
pothos, line 43:
```suggestion
It might be required to add a user different from `core` user to the `systemd-journal` group to read the user journal. It can be done with this Butane config:
```

pothos (approved):

docs/provisioning/config-transpiler/examples.md
```diff
@@ -190,6 +193,9 @@ storage:
 
           [Install]
           WantedBy=default.target
+    - path: /etc/userdb/flatcar:systemd-journal.membership
```
pothos, line 196: Since we had to add `systemd-journal` as regular group to make tmpfiles work for `/var/log/journal`, I think we can also use the normal Ignition group adding in this example.

docs/setup/debug/reading-the-system-log.md
```diff
@@ -38,6 +38,24 @@ Dec 22 12:32:39 localhost docker[9772]: /usr/sbin/apache2ctl: 87: ulimit: error
...
+It might be required to add a user different from `core` user to the `systemd-journal` group to read the user journal. It can be done with this Butane config:
+```yaml
+variant: flatcar
+version: 1.0.0
+passwd:
+  users:
+    - name: flatcar
+storage:
+  files:
+    - path: /etc/userdb/flatcar:systemd-journal.membership
```
pothos, line 52: (Same here)

pothos (approved):

PR #299, opened 2023-04-11
Consistent use of Butane vs Container Linux Config

pothos:
While many parts were already converted to Butane, the text sometimes used "Container Linux Config" and not all content was converted. While checking I also noticed some small mistakes that were corrected. The remaining use of Container Linux Config is mostly intentional, only the Terraform examples are to be converted in the near future but since they are self-contained, this is not urgent or inconsistent. The use of CLC for GRUB configuration is also kept as alternative to the kernel args setting in Butane when Ignition 2 is required.

## Testing done

Booted the custom Docker example and the `oem-postinst` example. The rest was checked by transpiling.


docs/setup/customization/customize-etcd-unit.md
```diff
@@ -11,23 +11,35 @@ The etcd systemd unit can be customized by overriding the unit that ships with t
 
 ## Use client certificates
 
-etcd supports client certificates as a way to provide secure communication between clients &#8596; leader and internal traffic between etcd peers in the cluster. Configuring certificates for both scenarios is done through the etcd section in a Container Linux Config. Options provided here will augment the unit that ships with Flatcar Container Linux.
+etcd supports client certificates as a way to provide secure communication between clients &#8596; leader and internal traffic between etcd peers in the cluster. Configuring certificates for both scenarios are done through a Butane Config. Options provided here will augment the unit that ships with Flatcar Container Linux.
```
krnowak, line 14: It should be `is` because it's about "configuring".

```suggestion
etcd supports client certificates as a way to provide secure communication between clients &#8596; leader and internal traffic between etcd peers in the cluster. Configuring certificates for both scenarios is done through a Butane Config. Options provided here will augment the unit that ships with Flatcar Container Linux.
```

t-lo:
There's some overlap of this PR and https://github.com/flatcar/flatcar-docs/pull/300; I'd propose to first get https://github.com/flatcar/flatcar-docs/pull/300 merged, rebase this PR, and address the comments above.

t-lo:
Rebased on latest main after #300 got merged.

docs/installing/cloud/azure.md
```diff
@@ -568,8 +568,6 @@ For each machine in the list, you should have a `machine-NAME.yaml.tmpl` file wi
 Create the configuration for `mynode` in the file `cl/machine-mynode.yaml.tmpl`:
 
 ```yaml
-variant: flatcar
-version: 1.0.0
```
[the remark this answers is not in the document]
t-lo: Could you please elaborate? I cant't find any mention of this in the previous version of that document...

Also, I am *entirely* unsure why we would *remove* this:
```yaml
variant: flatcar
version: 1.0.0
```
Afaikt this is *required* to make the Butane terraform provider work; see examples at https://github.com/poseidon/terraform-provider-ct.

t-lo:
@tormath1 Could you please have a look at the terraform examples which explicitly remove this:
```yaml
variant: flatcar
version: 1.0.0
```
Afaikt this is *required* to make the Butane terraform provider work; see examples at https://github.com/poseidon/terraform-provider-ct.

Your guidance would be greatly appreciated!

docs/installing/cloud/azure.md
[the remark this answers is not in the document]
t-lo: Or maybe I'm just missing where this is set in terraform code so it's not necessary when rendering the template?

docs/installing/cloud/openstack.md
```diff
@@ -191,10 +194,6 @@ nova boot \
 --security-groups default,flatcar
 ```
 
-## Multiple clusters
-
-If you would like to create multiple clusters you'll need to generate and use a new discovery token. Change the token value on the etcd discovery parameter in the Container Linux Config, and boot new instances.
-
```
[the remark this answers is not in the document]
t-lo: cl-config has a number of service-specific configuration fields, e.g. "etcd", "flannel", "docker", "update", and "locksmith". Butane supports none of these; our docs have been updated at the corresponding places to use plain config files (see e.g. setting locksmith reboot strategies here: https://www.flatcar.org/docs/latest/setup/releases/update-strategies/#reboot-strategy-options-through-butaneignition).

t-lo:
> @t-lo thanks for taking over this PR. You are right, `butane` needs to have these fields:
> 
> ```yaml
> variant: flatcar
> version: 1.0.0
> ```
> 
> but in these particular files (`equinix-metal.md` and `azure.md`), it's ok to keep CLC as these configs belong to the TF examples as @pothos mentioned:
> 
> > The remaining use of Container Linux Config is mostly intentional, only the Terraform examples are to be converted in the near future but since they are self-contained, this is not urgent or inconsistent.

But won't terraform fail to render these with an up-to-date `ct` provider?

t-lo:
Oh wait, I found it:
```terraform
      source  = "poseidon/ct"
      version = "0.7.1"      
```
Okay, that is a _really_ old version, we should fix that eventually.

But now I understand the removal.

docs/installing/cloud/azure.md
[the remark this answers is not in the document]
t-lo: @krnowak  Couldn't find it in the current documentation - could you please elaborate on what's missing, and why?

docs/installing/cloud/openstack.md
```diff
@@ -116,10 +118,11 @@ systemd:
 Transpile it to Ignition JSON:
 
 ```shell
-cat cl.yaml | docker run --rm -i ghcr.io/flatcar/ct:latest -platform openstack-metadata > ignition.json
+cat cl.yaml | docker run --rm -i quay.io/coreos/butane:release -platform openstack-metadata > ignition.json
```
krnowak, line 121: I think that there is no platform parameter in butane:

```suggestion
cat cl.yaml | docker run --rm -i quay.io/coreos/butane:release > ignition.json
```

docs/installing/cloud/vmware.md
```diff
@@ -87,35 +87,38 @@ Run VMware Workstation GUI:
 
 Flatcar Container Linux can also be installed by booting the virtual machine via [PXE][PXE] or the [ISO image][ISO] and then [installing Flatcar Container Linux to disk][install].
 
-## Container Linux Configs
+## Butane Configs
 
-Flatcar Container Linux allows you to configure machine parameters, configure networking, launch systemd units on startup, and more via Container Linux Configs (CLC). These configs are then [transpiled][transpiler] into Ignition configs and given to booting machines. Head over to the [docs to learn about the supported features][cl-configs].
+Flatcar Container Linux allows you to configure machine parameters, configure networking, launch systemd units on startup, and more via Butane Configs. These configs are then [transpiled][transpiler] into Ignition configs and given to booting machines. Head over to the [docs to learn about the supported features][transpiler].
```
krnowak, line 92: We have two links in the same sentence pointing to the same place. I'd drop one.

```suggestion
Flatcar Container Linux allows you to configure machine parameters, configure networking, launch systemd units on startup, and more via Butane Configs. These configs are then transpiled into Ignition configs and given to booting machines. Head over to the [docs to learn about the supported features][transpiler].
```

docs/installing/cloud/vmware.md
```diff
@@ -139,7 +142,7 @@ systemd:
 Transpile it to Ignition JSON:
 
 ```shell
-cat cl.yaml | docker run --rm -i ghcr.io/flatcar/ct:latest -platform custom > ignition.json
+cat cl.yaml | docker run --rm -i quay.io/coreos/butane:release -platform custom > ignition.json
```
krnowak, line 145: Same here about butane and platform parameter:
```suggestion
cat cl.yaml | docker run --rm -i quay.io/coreos/butane:release > ignition.json
```

docs/installing/community-platforms/notes-for-distributors.md
```diff
@@ -42,22 +42,23 @@ There are two predominant ways that a Flatcar Container Linux image can be easil
...
 
-Additionally, it is recommended that providers ensure that [coreos-metadata][coreos-metadata] and [butane][butane] have support for their platform. This will allow a nicer user experience, as coreos-metadata will be able to install users' ssh keys and users will be able to reference dynamic data in their Butane Configs.
+Additionally, it is recommended that providers ensure that [Afterburn][coreos-metadata] has support for their platform. This will allow a nicer user experience, as Afterburn will be able to install users' ssh keys and users will be able to reference metadata variables in their systemd units.
 
 [ignition]: ../../provisioning/ignition
 [ign-platforms]: https://github.com/coreos/ignition/blob/main/docs/supported-platforms.md
+[ign-platforms]: https://github.com/kinvolk/ignition/blob/master/doc/supported-platforms.md
```
krnowak, line 49: Duplicated `[ign-platforms]` link. And both are wrong - should be pointing to coreos repo.

docs/installing/community-platforms/notes-for-distributors.md
```diff
@@ -42,22 +42,23 @@ There are two predominant ways that a Flatcar Container Linux image can be easil
...
 [coreos-metadata]: https://github.com/coreos/afterburn/
-[butane]: https://github.com/coreos/butane
 
 ### Cloud config
 
-A Flatcar Container Linux image can also be customized using [cloud-config](https://github.com/flatcar/coreos-cloudinit/blob/master/Documentation/cloud-config.md). Users are recommended to instead use Butane Configs (that are converted into Ignition configs with [butane][butane]), for reasons [outlined in the blog post that introduced Ignition][ignition-blog].
+A Flatcar Container Linux image can also be customized using [cloud-config](https://github.com/kinvolk/coreos-cloudinit/blob/master/Documentation/cloud-config.md). Users are recommended to instead use Butane Configs (that are converted into Ignition configs with [Butane][butane-configs]), for reasons [outlined in the blog post that introduced Ignition][ignition-blog].
```
krnowak, line 54: Looks like a rebase gone awry here. Also, not sure what is the convention. For configuration written with butane we write "Butane Config(uration)" (or maybe "Butane config(uration)", what about the butane tool itself? Should it be "Butane" or "butane"? Or even `butane`?

```suggestion
A Flatcar Container Linux image can also be customized using [cloud-config](https://github.com/flatcar/coreos-cloudinit/blob/master/Documentation/cloud-config.md). Users are recommended to instead use Butane Configs (that are converted into Ignition configs with [`butane`][butane-configs]), for reasons [outlined in the blog post that introduced Ignition][ignition-blog].
```

docs/installing/vms/libvirt.md
```diff
@@ -107,13 +107,14 @@ echo "  # For ignition files" >> /etc/apparmor.d/abstractions/libvirt-qemu
 echo "  /var/lib/libvirt/flatcar-linux/** r," >> /etc/apparmor.d/abstractions/libvirt-qemu
 ```
 
-Since the empty Ignition config is not very useful, here is an example how to write a simple Flatcar Container Linux config to add your ssh keys and write a hostname file:
+Since the empty Ignition config is not very useful, here is an example how to write a simple Butane config to add your ssh keys and write a hostname file:
```
krnowak, line 110:
```suggestion
Since the empty Ignition config is not very useful, here is an example how to write a simple Butane Config to add your ssh keys and write a hostname file:
```

docs/installing/vms/libvirt.md
```diff
@@ -168,9 +169,11 @@ ssh core@192.168.122.184
 
 #### Static IP
 
-By default, Flatcar Container Linux uses DHCP to get its network configuration. In this example the VM will be attached directly to the local network via a bridge on the host's virbr0 and the local network. To configure a static address add a [networkd unit][systemd-network] to the Flatcar Container Linux config:
+By default, Flatcar Container Linux uses DHCP to get its network configuration. In this example the VM will be attached directly to the local network via a bridge on the host's virbr0 and the local network. To configure a static address add a [networkd unit][systemd-network] to the Butane config:
```
krnowak, line 172:
```suggestion
By default, Flatcar Container Linux uses DHCP to get its network configuration. In this example the VM will be attached directly to the local network via a bridge on the host's virbr0 and the local network. To configure a static address add a [networkd unit][systemd-network] to the Butane Config:
```

docs/installing/cloud/azure.md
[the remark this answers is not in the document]
t-lo: Oh, today I learned! That's what I get from only using Butane / Ignition 3.0 😅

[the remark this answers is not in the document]
t-lo: Fixed.

docs/installing/cloud/digitalocean.md
```diff
@@ -349,8 +349,6 @@ For example, create the configuration for `mynode` in the file `machine-mynode.y
 
 ```yaml
 ---
-variant: flatcar
-version: 1.0.0
```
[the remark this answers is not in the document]
t-lo: Done, good catch.

docs/installing/cloud/equinix-metal.md
```diff
@@ -276,9 +276,6 @@ ssh_keys     = ["ssh-rsa AA... me@mail.net"]
 Create the configuration for `mynode` in the file `machine-mynode.yaml.tmpl`:
 
 ```yaml
----
-variant: flatcar
-version: 1.0.0
```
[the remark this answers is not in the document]
t-lo: Fixed.

docs/installing/community-platforms/notes-for-distributors.md
t-lo: No idea how that slipped through my search+replace. Good catch, will fix.

t-lo:
I've updated the Hetzner terraform code to match https://github.com/flatcar/flatcar-terraform/blob/main/flatcar-terraform-hetzner/README.md - it's the first terraform config we ported to Ignition 3.0 / Butane.

docs/installing/cloud/hetzner.md
```diff
@@ -158,63 +160,167 @@ variable "cluster_name" {
...
+server_type = "cx11"
+
+# Additional SSH "authorized hosts" keys for the "core" user.
+# ssh_keys = [ "...", "..." ]
+
+# One of "lts", "stable", "beta", or "alpha"
+release_channel = "stable"
+```
+
+The above references a deployment configuration in [Butane](https://www.flatcar.org/docs/latest/provisioning/config-transpiler/configuration/) syntax; `server-configs/server1.yaml`.
```
krnowak, line 294: I'm not sure we want to use absolute paths here. Should it be something like

```suggestion
The above references a deployment configuration in [Butane](../../../provisioning/config-transpiler/configuration/) syntax; `server-configs/server1.yaml`.
```

krnowak (approved):

PR #301, opened 2023-04-20
container-runtimes/docker: remove docker.socket deactivation

tormath1:
This socket deactivation is useless (and never worked since a while) as the socket is anyway always activated by Torcx.

Enabling the Docker service is enough to have container restarting at boot.

<hr>

Related to: https://github.com/flatcar/Flatcar/issues/1004

### DIFF
--- a/docs/container-runtimes/getting-started-with-docker.md
+++ b/docs/container-runtimes/getting-started-with-docker.md
@@ -102,9 +102,7 @@ Here is a Container Linux Config to enable the Docker service while disabling so
 ```yaml
 systemd:
   units:
-    # Ensure docker starts automatically instead of being socket-activated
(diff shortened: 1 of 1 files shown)


docs/container-runtimes/getting-started-with-docker.md
```diff
@@ -103,8 +103,6 @@ Here is a Container Linux Config to enable the Docker service while disabling so
 systemd:
   units:
     # Ensure docker starts automatically instead of being socket-activated
```
krnowak, line 105: We could update this comment then.
```suggestion
    # Ensure docker starts automatically instead of being only socket-activated
```

krnowak (approved):

PR #302, opened 2023-04-24
Bump hetznercloud/hcloud provider to v1.38.2 version 

ThisIsQasim:


t-lo (approved):
Follow-up from https://github.com/flatcar/flatcar-terraform/pull/15.

PR #313, opened 2023-06-01
Update architectures.md to fix https://github.com/flatcar/Flatcar/iss…

krishjainx:
# Update architectures.md to fix [https://github.com/flatcar/Flatcar/iss…](https://github.com/flatcar/Flatcar/issues/260)

This pull request addresses the warning message that appears when using the current documentation and Butane config. The warning states that the 'enable-docker-tcp.service' unit is enabled but lacks an install section, rendering the enable command ineffective. The proposed changes in this pull request resolve this issue.

# Testing

image

### DIFF
--- a/docs/setup/clusters/architectures.md
+++ b/docs/setup/clusters/architectures.md
@@ -36,7 +36,6 @@ systemd:
   units:
     - name: docker-tcp.socket
       enabled: true
-      mask: false
       contents: |
         [Unit]
         Description=Docker Socket for the API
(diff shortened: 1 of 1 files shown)


docs/setup/clusters/architectures.md
```diff
@@ -56,6 +56,8 @@ systemd:
         [Service]
         Type=oneshot
         ExecStart=/usr/bin/systemctl enable docker-tcp.socket
+        [Install]
+        WantedBy=multi-user.target
```
pothos, line 60: I don't understand why `enable-docker-tcp.service` even exists. Can't we remove this unit? The `docker-tcp.socket` unit is enabled and created above (I also don't get why there has to be a `mask: false` entry).

pothos: Did you test the instructions?

pothos: Thanks

pothos (approved):

PR #326, opened 2023-07-11
update-strategies: add missing "overwrite: true"

t-lo:
Our butane examples for customising locksmith contain an error: `overwrite: true` is missing when modifying `update.conf`. This leads to ignition panicking during provisioning as `update.conf` already exists:

```
ignition[1009]: CRITICAL : Ignition failed: failed to create files: failed to create files: error creating /sysroot/etc/flatcar/update.conf: error creating file "/sysroot/etc/flatcar/update.conf": A file exists there already and overwrite is false
```

This change updates our examples and adds `overwrite:true`.

### DIFF
--- a/docs/setup/releases/update-strategies.md
+++ b/docs/setup/releases/update-strategies.md
@@ -56,6 +56,7 @@ version: 1.0.0
 storage:
   files:
     - path: /etc/flatcar/update.conf
+      overwrite: true
       contents:
         inline: |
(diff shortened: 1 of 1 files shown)


pothos:
Thanks, missed that in the migration (the default Ignition behavior switched).

By the way, there is also `append:` to extend the existing file, might be useful now, too, for some cases where a drop-in is not possible.

PR #340, opened 2023-10-05
Skip the unpacking of the QEMU image

pothos:
We now provide the qcow2 image with inline compression. This means we can skip the unpacking. This is available from Stable 3602.2.0 and the hands-on 4 tutorial was updated to this version - hence we should wait for 3602.2.1 to be out before merging this change.

## Testing done

None


tormath1:
Hello, this repo is now archived. This PR should be re-opened against: https://github.com/flatcar/flatcar-website.

To update your commit(s):
```
PR_ID=340
cd flatcar/flatcar-website
wget "https://patch-diff.githubusercontent.com/raw/flatcar/flatcar-docs/pull/${PR_ID}.patch"
sed -i 's|/docs/|/content/docs/latest/|g' "${PR_ID}.patch"
git am "${PR_ID}.patch"
```

